PR #17, opened 2022-12-12
Reorganize events due to database migration

natalian98:
Summary of changes:
- only run migration on database created event
- add an action for migration
- do not start the service by default (`"startup": "disabled"`) as it should be started only once database details required to execute the sql migration are in place
- no need to check for postgres relation on database created - if this event was emitted, we're assuming the relation is in place
- removed config file comparison and pushing it on first layer update - the file will be pushed if database was created or changed
- added event handling for postgres relation departed (since there is no support for database_removed in `database_requires`)
- reorganized unit tests
- changed database name from `postgres` to app name
- pass ports as variables
- upgrade image to v2.0.3

### DIFF
--- a/actions.yaml
+++ b/actions.yaml
@@ -1,0 +1,2 @@
+run-migration:
+  description: Run a command to create SQL schemas and apply migration plans.
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -18,3 +18,7 @@ requires:
   pg-database:
     interface: postgresql_client
     limit: 1
+  public-ingress:
+    interface: ingress
+  admin-ingress:
+    interface: ingress
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -13,7 +13,7 @@ resources:
   oci-image:
     type: oci-image
     description: OCI image for hydra container
-    upstream-source: oryd/hydra:v2.0.2
+    upstream-source: oryd/hydra:v2.0.3
 requires:
   pg-database:
     interface: postgresql_client
--- a/src/charm.py
+++ b/src/charm.py
@@ -11,7 +11,7 @@ import logging
 import yaml
 from charms.data_platform_libs.v0.database_requires import DatabaseRequires
 from charms.observability_libs.v0.kubernetes_service_patch import KubernetesServicePatch
-from ops.charm import CharmBase
+from ops.charm import ActionEvent, CharmBase
 from ops.main import main
 from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, ModelError, WaitingStatus
 from ops.pebble import ChangeError, ExecError, Layer
@@ -48,6 +48,7 @@ class HydraCharm(CharmBase):
         self.framework.observe(self.on.hydra_pebble_ready, self._on_hydra_pebble_ready)
         self.framework.observe(self.database.on.database_created, self._on_database_created)
         self.framework.observe(self.database.on.endpoints_changed, self._on_database_changed)
+        self.framework.observe(self.on.run_migration_action, self._on_run_migration)
         self.framework.observe(
             self.on["pg-database"].relation_departed, self._on_database_relation_departed
         )
@@ -63,7 +64,9 @@ class HydraCharm(CharmBase):
                     "override": "replace",
                     "summary": "entrypoint of the hydra-operator image",
                     "command": f"hydra serve all --config {self._hydra_config_path} --dev",
-                    "startup": "disabled",
+                    "startup": "disabled"
+                    if not self.database.is_database_created()
+                    else "enabled",
                 }
             },
             "checks": {
@@ -126,8 +129,7 @@ class HydraCharm(CharmBase):
     def _run_sql_migration(self) -> None:
         """Runs a command to create SQL schemas and apply migration plans."""
         process = self._container.exec(
-            ["hydra", "migrate", "sql", "-e", "--config", self._hydra_config_path, "--yes"],
-            timeout=20.0,
+            ["hydra", "migrate", "sql", "-e", "--config", self._hydra_config_path, "--yes"]
         )
         try:
             stdout, _ = process.wait_output()
@@ -135,6 +137,7 @@ class HydraCharm(CharmBase):
         except ExecError as err:
             logger.error(f"Exited with code {err.exit_code}. Stderr: {err.stderr}")
             self.unit.status = BlockedStatus("Database migration job failed")
+            logger.error("Automigration job failed, please use the run-migration action")
 
     def _on_hydra_pebble_ready(self, event) -> None:
         """Event Handler for pebble ready event."""
@@ -161,11 +164,15 @@ class HydraCharm(CharmBase):
             logger.error(str(err))
             self.unit.status = BlockedStatus("Failed to replan")
             return
-
-        if self.database.is_database_created():
-            self._container.push(self._hydra_config_path, self._config, make_dirs=True)
-            self._container.start(self._container_name)
-            self.unit.status = ActiveStatus()
+        finally:
+            if self.database.is_database_created():
+                if not self._container.exists(self._hydra_config_path):
+                    logger.info(
+                        "Config file not found. Pushing it to the container and replanning"
+                    )
+                    self._container.push(self._hydra_config_path, self._config, make_dirs=True)
+                    self._container.replan()
+                self.unit.status = ActiveStatus()
 
     def _on_database_created(self, event) -> None:
         """Event Handler for database created event."""
@@ -188,6 +195,7 @@ class HydraCharm(CharmBase):
         )
 
         try:
+            self._container.add_layer(self._container_name, self._hydra_layer, combine=True)
             self._container.get_service(self._container_name)
             logger.info("Updating Hydra config and restarting service")
             self._container.push(self._hydra_config_path, self._config, make_dirs=True)
@@ -219,6 +227,11 @@ class HydraCharm(CharmBase):
             event.defer()
             self.unit.status = WaitingStatus("Waiting for Hydra service")
             logger.info("Hydra service is absent. Deferring database created event.")
+
+    def _on_run_migration(self, event: ActionEvent) -> None:
+        """Runs the migration as an action response."""
+        logger.info("Executing database migration initiated by user")
+        self._run_sql_migration()
 
     def _on_database_relation_departed(self, event) -> None:
         """Event Handler for database relation departed event."""
--- a/src/charm.py
+++ b/src/charm.py
@@ -140,11 +140,6 @@ class HydraCharm(CharmBase):
 
     def _on_hydra_pebble_ready(self, event) -> None:
         """Event Handler for pebble ready event."""
-        if not self.model.relations[self._db_relation_name]:
-            logger.error("Missing required relation with postgresql")
-            self.model.unit.status = BlockedStatus("Missing required relation with postgresql")
-            return
-
         if not self._container.can_connect():
             event.defer()
             logger.info("Cannot connect to Hydra container. Deferring the event.")
@@ -167,8 +162,12 @@ class HydraCharm(CharmBase):
             self._container.push(self._hydra_config_path, self._config, make_dirs=True)
             self._container.start(self._container_name)
             self.unit.status = ActiveStatus()
-        else:
+            return
+
+        if self.model.relations[self._db_relation_name]:
             self.unit.status = WaitingStatus("Waiting for database creation")
+        else:
+            self.unit.status = BlockedStatus("Missing required relation with postgresql")
 
     def _on_database_created(self, event) -> None:
         """Event Handler for database created event."""
--- a/src/charm.py
+++ b/src/charm.py
@@ -33,6 +33,7 @@ class HydraCharm(CharmBase):
         self._container = self.unit.get_container(self._container_name)
         self._hydra_config_path = "/etc/config/hydra.yaml"
         self._name = self.model.app.name
+        self._db_relation_name = "pg-database"
 
         self.service_patcher = KubernetesServicePatch(
             self, [("hydra-admin", HYDRA_ADMIN_PORT), ("hydra-public", HYDRA_PUBLIC_PORT)]
@@ -40,7 +41,7 @@ class HydraCharm(CharmBase):
 
         self.database = DatabaseRequires(
             self,
-            relation_name="pg-database",
+            relation_name=self._db_relation_name,
             database_name=f"{self.model.name}_{self._name}",
             extra_user_roles=EXTRA_USER_ROLES,
         )
@@ -50,7 +51,7 @@ class HydraCharm(CharmBase):
         self.framework.observe(self.database.on.endpoints_changed, self._on_database_changed)
         self.framework.observe(self.on.run_migration_action, self._on_run_migration)
         self.framework.observe(
-            self.on["pg-database"].relation_departed, self._on_database_relation_departed
+            self.on[self._db_relation_name].relation_departed, self._on_database_relation_departed
         )
 
     @property
@@ -139,7 +140,7 @@ class HydraCharm(CharmBase):
 
     def _on_hydra_pebble_ready(self, event) -> None:
         """Event Handler for pebble ready event."""
-        if not self.model.relations["pg-database"]:
+        if not self.model.relations[self._db_relation_name]:
             logger.error("Missing required relation with postgresql")
             self.model.unit.status = BlockedStatus("Missing required relation with postgresql")
             return
@@ -159,14 +160,15 @@ class HydraCharm(CharmBase):
             self._container.replan()
         except ChangeError as err:
             logger.error(str(err))
-            self.unit.status = BlockedStatus("Failed to replan")
+            self.unit.status = BlockedStatus("Failed to replan, please consult the logs")
             return
 
         if self.database.is_database_created():
             self._container.push(self._hydra_config_path, self._config, make_dirs=True)
             self._container.start(self._container_name)
             self.unit.status = ActiveStatus()
-            return
+        else:
+            self.unit.status = WaitingStatus("Waiting for database creation")
 
     def _on_database_created(self, event) -> None:
         """Event Handler for database created event."""
--- a/src/charm.py
+++ b/src/charm.py
@@ -9,9 +9,13 @@
 import logging
 
 import yaml
-from charms.data_platform_libs.v0.database_requires import DatabaseRequires
+from charms.data_platform_libs.v0.database_requires import (
+    DatabaseCreatedEvent,
+    DatabaseEndpointsChangedEvent,
+    DatabaseRequires,
+)
 from charms.observability_libs.v0.kubernetes_service_patch import KubernetesServicePatch
-from ops.charm import ActionEvent, CharmBase
+from ops.charm import ActionEvent, CharmBase, RelationDepartedEvent, WorkloadEvent
 from ops.main import main
 from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, ModelError, WaitingStatus
 from ops.pebble import ChangeError, ExecError, Layer
@@ -128,7 +132,7 @@ class HydraCharm(CharmBase):
             "endpoints": relation_data.get("endpoints"),
         }
 
-    def _run_sql_migration(self, set_timeout) -> None:
+    def _run_sql_migration(self, set_timeout: bool) -> None:
         """Runs a command to create SQL schemas and apply migration plans."""
         process = self._container.exec(
             ["hydra", "migrate", "sql", "-e", "--config", self._hydra_config_path, "--yes"],
@@ -138,7 +142,7 @@ class HydraCharm(CharmBase):
         stdout, _ = process.wait_output()
         logger.info(f"Executing automigration: {stdout}")
 
-    def _on_hydra_pebble_ready(self, event) -> None:
+    def _on_hydra_pebble_ready(self, event: WorkloadEvent) -> None:
         """Event Handler for pebble ready event."""
         if not self._container.can_connect():
             event.defer()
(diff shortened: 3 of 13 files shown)


src/charm.py
```diff
@@ -121,92 +123,107 @@ def _get_database_relation_info(self) -> dict:
...
-                self._container.restart()
-
-    def _update_container(self, event) -> None:
-        """Update configs, pebble layer and run database migration."""
-        self.unit.status = MaintenanceStatus("Configuring hydra container")
+            stdout, _ = process.wait_output()
+            logger.info(f"Executing automigration: {stdout}")
+        except ExecError as err:
+            logger.error(f"Exited with code {err.exit_code}. Stderr: {err.stderr}")
+            self.unit.status = BlockedStatus("Database migration job failed")
```
nsklikas, line 137: What happens if this times out? 

We need to also have an action for running migrations, maybe in case of timeout (or all errors?) we could print a message telling the user to run the migration action

src/charm.py
```diff
@@ -121,92 +123,107 @@ def _get_database_relation_info(self) -> dict:
...
             self.unit.status = WaitingStatus("Waiting to connect to Hydra container")
             return
 
-        self._update_layer()
+        self.unit.status = MaintenanceStatus("Configuring resources")
+
+        self._container.add_layer(self._container_name, self._hydra_layer, combine=True)
+        logger.info("Pebble plan updated with new configuration, replanning")
+        try:
+            self._container.replan()
```
nsklikas, line 159: The pebble layer will always have `"startup": "disabled"`, this means that this call will always do nothing

src/charm.py
```diff
@@ -121,92 +123,107 @@ def _get_database_relation_info(self) -> dict:
...
+            self.unit.status = ActiveStatus()
+
+    def _on_database_created(self, event) -> None:
+        """Event Handler for database created event."""
+        logger.info("Retrieved database details")
 
         if not self.unit.is_leader():
             # TODO: Observe leader_elected event
             logger.info("Unit does not have leadership")
             self.unit.status = WaitingStatus("Waiting for leadership")
```
nsklikas, line 177: If a unit is not the leader, it does not mean that there is no leader. You should not change the status of the unit here.

src/charm.py
```diff
@@ -61,7 +63,7 @@ def _hydra_layer(self) -> Layer:
                     "override": "replace",
                     "summary": "entrypoint of the hydra-operator image",
                     "command": f"hydra serve all --config {self._hydra_config_path} --dev",
-                    "startup": "enabled",
+                    "startup": "disabled",
```
nsklikas, line 66: how about:
```suggestion
                    "startup": "disabled" if not self.database.is_database_created() else "enabled",
```

src/charm.py
```diff
@@ -121,92 +123,107 @@ def _get_database_relation_info(self) -> dict:
...
-            stdout, _ = process.wait_output()
-            logger.info(f"Executing automigration: {stdout}")
-        except ExecError as err:
-            logger.error(f"Exited with code {err.exit_code}. Stderr: {err.stderr}")
-            self.unit.status = BlockedStatus("Database migration job failed")
+            self._container.get_service(self._container_name)
+            logger.info("Updating Hydra config and restarting service")
+            self._container.push(self._hydra_config_path, self._config, make_dirs=True)
+            self._run_sql_migration()
+            self._container.restart(self._container_name)
```
nsklikas, line 195: We could push the pebble layer again here and update it to `"startup": "enabled"`

src/charm.py
```diff
@@ -121,92 +123,107 @@ def _get_database_relation_info(self) -> dict:
...
+            self._container.replan()
+        except ChangeError as err:
+            logger.error(str(err))
+            self.unit.status = BlockedStatus("Failed to replan")
+            return
+
+        if self.database.is_database_created():
+            self._container.push(self._hydra_config_path, self._config, make_dirs=True)
+            self._container.start(self._container_name)
+            self.unit.status = ActiveStatus()
```
nsklikas, line 168: We are also doing this in `_on_database_created`, do we need to do it again here?

nsklikas (changes requested):
Good job overall, looks a lot better than before. I left a couple of comments. 

We should have an action for running migrations. Would you like to add it in this PR?

src/charm.py
natalian98: The migration shouldn't really take that long, but I removed the 20s timeout. Errors should be caught by `ops.pebble.ExecError`. I added an action to run the migration manually.

src/charm.py
natalian98: I'm not sure I understand. The waiting status is set only to non-leader unit, not the entire charm. Is that wrong?

nsklikas: The message is wrong. The unit is not waiting for leadership, it is waiting for the leader to run the migration. I am fine with changing the status to waiting, but we should have a better message for it.

src/charm.py
```diff
@@ -83,7 +88,7 @@ def _config(self) -> str:
         db_info = self._get_database_relation_info()
 
         config = {
-            "dsn": f"postgres://{db_info['username']}:{db_info['password']}@{db_info['endpoints']}/postgres",
+            "dsn": f"postgres://{db_info['username']}:{db_info['password']}@{db_info['endpoints']}/{self._name}",
```
nsklikas, line 91: I am not sure if there will ever be a problem with this, but I think that `{self.model.name}_{self._name}` would be safer here (to avoid collisions with apps of the same name running in different models.

src/charm.py
```diff
@@ -121,92 +126,117 @@ def _get_database_relation_info(self) -> dict:
...
+        except ExecError as err:
+            logger.error(f"Exited with code {err.exit_code}. Stderr: {err.stderr}")
+            self.unit.status = BlockedStatus("Database migration job failed")
+            logger.error("Automigration job failed, please use the run-migration action")
 
+    def _on_hydra_pebble_ready(self, event) -> None:
+        """Event Handler for pebble ready event."""
         if not self.model.relations["pg-database"]:
             # TODO: Observe relation_joined event instead of event deferring
             event.defer()
```
nsklikas, line 146: Deferring is not needed here. The service will be started by the `_on_database_created` hook.

src/charm.py
nsklikas: But do you know if the config file has changed or not? Maybe we can push the config file every time.

src/charm.py
nsklikas: How long the migration takes depends on the volume of the data in the database and on the migration commands. If the migration runs from a hook we should have a timeout set, if the user runs the migration we can wait it out.

src/charm.py
nsklikas, line 146: Deferring is not needed here. The service will be started by the `_on_database_created` hook.

src/charm.py
natalian98: Ok, message changed. Thanks for explanation

src/charm.py
natalian98: We don't need to check if the config was changed. If it does, it will be pushed `_on_database_changed` (for now the only thing that could change in the config is the endpoint).
So there's no need to push this file it every time.

src/charm.py
natalian98: Good point, thanks. I removed it

src/charm.py
```diff
@@ -121,92 +126,116 @@ def _get_database_relation_info(self) -> dict:
...
+
+    def _on_run_migration(self, event: ActionEvent) -> None:
+        """Runs the migration as an action response."""
+        logger.info("Executing database migration initiated by user")
+        self._run_sql_migration(set_timeout=False)
+
+    def _on_database_relation_departed(self, event) -> None:
+        """Event Handler for database relation departed event."""
+        logger.error("Missing required relation with postgresql")
+        self.model.unit.status = BlockedStatus("Missing required relation with postgresql")
```
nsklikas, line 238: Should we update the Hydra config here (ie remove the `dsn` or stop the service). I assume that a departed relation means that the database will be gone and that Hydra will be trying to call the database APIs and get errors.

src/charm.py
```diff
@@ -121,92 +126,116 @@ def _get_database_relation_info(self) -> dict:
...
             return
+        finally:
+            if self.database.is_database_created():
+                if not self._container.exists(self._hydra_config_path):
+                    logger.info(
+                        "Config file not found. Pushing it to the container and replanning"
+                    )
+                    self._container.push(self._hydra_config_path, self._config, make_dirs=True)
+                    self._container.replan()
+                self.unit.status = ActiveStatus()
```
nsklikas, line 174: This could be rewritten as:
```suggestion
        self._container.push(self._hydra_config_path, self._config, make_dirs=True)
        try:
            self._container.replan()
        except ChangeError as err:
            logger.error(str(err))
            self.unit.status = BlockedStatus("Failed to replan")
            return
        self.unit.status = ActiveStatus()
```
This way we are pushing the config file every time (which is not needed), but the flow is much simpler. What do you think?

src/charm.py
```diff
@@ -121,92 +126,116 @@ def _get_database_relation_info(self) -> dict:
...
+        )
+
+        try:
+            self._container.add_layer(self._container_name, self._hydra_layer, combine=True)
+            self._container.get_service(self._container_name)
+            logger.info("Updating Hydra config and restarting service")
+            self._container.push(self._hydra_config_path, self._config, make_dirs=True)
+            self._run_sql_migration(set_timeout=True)
+            self._container.restart(self._container_name)
+            self.unit.status = ActiveStatus()
```
nsklikas, line 203: I think that the only line that can throw is:
```python
self._container.restart(self._container_name)
```
Can we move the rest of the code out of the `try` block?

src/charm.py
```diff
@@ -121,92 +126,116 @@ def _get_database_relation_info(self) -> dict:
...
-
-    def _on_db_events(self, db_event) -> None:
-        """Event Handler for database-related events."""
-        logger.info("Retrieved database details")
-        self._update_container(db_event)
+            self._container.get_service(self._container_name)
+            logger.info("Updating Hydra config and restarting service")
+            self._container.push(self._hydra_config_path, self._config, make_dirs=True)
+            self._container.restart(self._container_name)
+            self.unit.status = ActiveStatus()
```
nsklikas, line 224: I think that the only line that can throw is:
```python
self._container.restart(self._container_name)
```
Can we move the rest of the code out of the `try` block?

nsklikas (commented):
LGTM, left some minor comments

src/charm.py
```diff
@@ -61,7 +64,9 @@ def _hydra_layer(self) -> Layer:
                     "override": "replace",
                     "summary": "entrypoint of the hydra-operator image",
                     "command": f"hydra serve all --config {self._hydra_config_path} --dev",
-                    "startup": "enabled",
+                    "startup": "disabled"
```
gruyaume, line 67: Why add the pebble layer on pebble ready when the database is not yet created. We could simply add the pebble on database created instead.

actions.yaml
```diff
@@ -0,0 +1,2 @@
+run-migration:
+  description: Run a command to create SQL schemas and apply migration plans.
```
gruyaume, line 2: In the future, please make the PR's simplers. Don't mix in new features with refactoring

metadata.yaml
```diff
@@ -13,7 +13,7 @@ resources:
   oci-image:
     type: oci-image
     description: OCI image for hydra container
-    upstream-source: oryd/hydra:v2.0.2
+    upstream-source: oryd/hydra:v2.0.3
```
gruyaume, line 16: In the future, please make the PR's simplers. Don't mix in new features with refactoring.

src/charm.py
```diff
@@ -121,92 +126,116 @@ def _get_database_relation_info(self) -> dict:
...
+            self.unit.status = WaitingStatus("Waiting to connect to Hydra container")
             return
 
-        if not self.database.is_database_created():
+        self.unit.status = MaintenanceStatus(
+            "Configuring container and resources for database connection"
+        )
+
+        try:
+            self._container.add_layer(self._container_name, self._hydra_layer, combine=True)
```
gruyaume, line 197: Why so many items in the try/except. It's not obvious which calls can lead to which error.

src/charm.py
```diff
@@ -121,92 +126,116 @@ def _get_database_relation_info(self) -> dict:
...
+        process = self._container.exec(
+            ["hydra", "migrate", "sql", "-e", "--config", self._hydra_config_path, "--yes"],
+            timeout=20.0 if set_timeout else None,
+        )
+        try:
+            stdout, _ = process.wait_output()
+            logger.info(f"Executing automigration: {stdout}")
+        except ExecError as err:
+            logger.error(f"Exited with code {err.exit_code}. Stderr: {err.stderr}")
+            self.unit.status = BlockedStatus("Database migration job failed")
```
gruyaume, line 140: Do we really want the outcome to be Blocked when the user runs the action manually? It should probably only be event.fail() in those cases. Error handling should probably be handled in the event methods and the generic method here should not set status.

gruyaume (commented):
As a general comment I'd recommend to not bundle new features (like the new action and the new image) with refactoring. The reason is now if the charm has problems it's hard to decipher why. It also makes it more complex to review.

src/charm.py
nsklikas: I see your point, but if the user tries to run a migration and that fails, doesn't it mean that the application is blocked? Assuming it's not a juju related error, if the migration fails to run it probably means that (all or some) hydra queries might fail.

Perhaps a better approach would be to set a flag with the latest migration that has been run in the peer relation (maybe use the app version?) and then use that in the `update_status` hook to set the app's status. 
e.g. `self.app.status = (migration_version == app_version)`

src/charm.py
natalian98: This line can also throw `ModelError` if the service doesn't exist:
```
self._container.get_service(self._container_name)
```
In that case the container shouldn't be restarted.

src/charm.py
natalian98: Thanks for your review @gruyaume.
I think the layer should be added on both `database_created` and `pebble_ready`. The reason is, a new database is created only once on install, but in case the container crashes only `pebble_ready` will be re-run, so the layer needs to be added here again.
In other reviews we also agreed to add a layer even if the database is not created. What do you think?

actions.yaml
natalian98: I'll try to keep the changes separate in future PRs. Adding the action was a part of addressing review comments. Thanks

src/charm.py
natalian98: With recent changes, if the migration run by charm fails (e.g. due to timeout or missing config file), the status will be set to blocked and suggest in logs that the action should be run instead. If this one fails as well, the event will fail but no status will be set. Let me know if that looks better now.

gruyaume: If the user runs the action too early for example, when the workload is not yet ready (let's say it's in maintenance), here we wouldn't want the action to place the charm in a blocked state but to simply fail and return a message like "can't complete the action because the workload is not running yet".

src/charm.py
```diff
@@ -116,19 +122,35 @@ def _get_database_relation_info(self) -> dict:
...
-        if not self._container.get_plan().to_dict():
-            # Push the config on first layer update to avoid PathError
-            self._container.push(self._hydra_config_path, self._config, make_dirs=True)
-            logger.info("Pushed hydra config")
+        stdout, _ = process.wait_output()
+        logger.info(f"Executing automigration: {stdout}")
+
+    def _on_hydra_pebble_ready(self, event) -> None:
+        """Event Handler for pebble ready event."""
+        if not self.model.relations["pg-database"]:
```
gruyaume, line 142: `pg-database` should probably a variable

src/charm.py
```diff
@@ -140,73 +162,97 @@ def _update_layer(self) -> None:
             self.unit.status = BlockedStatus("Failed to replan")
```
gruyaume, line 162: `BlockedStatus` should be used when we expect a human to do something, here it's not clear to me what they should do.

src/charm.py
```diff
@@ -116,19 +122,35 @@ def _get_database_relation_info(self) -> dict:
...
+        stdout, _ = process.wait_output()
+        logger.info(f"Executing automigration: {stdout}")
+
+    def _on_hydra_pebble_ready(self, event) -> None:
+        """Event Handler for pebble ready event."""
+        if not self.model.relations["pg-database"]:
+            logger.error("Missing required relation with postgresql")
+            self.model.unit.status = BlockedStatus("Missing required relation with postgresql")
+            return
+
```
gruyaume, line 146: Here should we have a condition that validates if the database is created and if not is placed in waiting state (`if not self.database.is_database_created`). This would mean that we 
1. Check if relation is created -> else block
2. Check if database is created -> else wait
3. Check if can connect to workload -> else wait

Then the rest (configuring pebble) is done without if/elses

natalian98: I am fine with that approach, but we agreed to follow a pattern in kratos and hydra to add the layer before checking if the database is created.
Is it fine to:
1. Check if relation is created -> else block
2. Check if can connect to workload -> else wait
3. _Add layer and replan_
4. Check if database is created -> else wait?

gruyaume: My concern is with the order of 3/4. It doesn't make much sense to me to create the layer before the database is created. Following this logic, why check if the relation with the db is created before adding the layer?

natalian98: I reordered it to check both the relation and database presence after the layer is added. I think it's fine to create it even before the database is created, as we only need to push a config file then.

src/charm.py
```diff
@@ -137,76 +155,105 @@ def _update_layer(self) -> None:
...
-        """Runs a command to create SQL schemas and apply migration plans."""
-        process = self._container.exec(
-            ["hydra", "migrate", "sql", "-e", "--config", self._hydra_config_path, "--yes"],
-            timeout=20.0,
-        )
+    def _on_run_migration(self, event: ActionEvent) -> None:
+        """Runs the migration as an action response."""
+        if not self.unit.is_leader():
+            event.fail("The action can be run only on leader unit")
+            return
```
nsklikas, line 241: If the user wants to run the migration from a non-leader unit, I see no reason to stop them.

nsklikas (changes requested):

src/charm.py
```diff
@@ -137,76 +155,105 @@ def _update_layer(self) -> None:
...
-        """Event Handler for pebble ready event."""
-        self._update_container(event)
+            event.fail("Execution failed, please inspect the logs")
+            return
 
-    def _on_db_events(self, db_event) -> None:
-        """Event Handler for database-related events."""
-        logger.info("Retrieved database details")
-        self._update_container(db_event)
+    def _on_database_relation_departed(self, event) -> None:
```
gruyaume, line 251: Add typing

src/charm.py
```diff
@@ -137,76 +155,105 @@ def _update_layer(self) -> None:
...
+        except ExecError as err:
+            logger.error(f"Exited with code {err.exit_code}. Stderr: {err.stderr}")
+            self.unit.status = BlockedStatus("Database migration job failed")
+            logger.error("Automigration job failed, please use the run-migration action")
             return
 
+        self._container.start(self._container_name)
+        self.unit.status = ActiveStatus()
+
+    def _on_database_changed(self, event) -> None:
```
gruyaume, line 214: Add typing

src/charm.py
```diff
@@ -137,76 +155,105 @@ def _update_layer(self) -> None:
...
+        if self.model.relations[self._db_relation_name]:
+            self.unit.status = WaitingStatus("Waiting for database creation")
         else:
-            if current_config != self._config:
-                self._container.push(self._hydra_config_path, self._config, make_dirs=True)
-                logger.info("Updated hydra config, restarting the container")
-                self._container.restart()
+            self.unit.status = BlockedStatus("Missing required relation with postgresql")
+
+    def _on_database_created(self, event) -> None:
```
gruyaume, line 172: Add typing

src/charm.py
```diff
@@ -116,19 +123,30 @@ def _get_database_relation_info(self) -> dict:
...
+        )
+
+        stdout, _ = process.wait_output()
+        logger.info(f"Executing automigration: {stdout}")
 
-        if not self._container.get_plan().to_dict():
-            # Push the config on first layer update to avoid PathError
-            self._container.push(self._hydra_config_path, self._config, make_dirs=True)
-            logger.info("Pushed hydra config")
+    def _on_hydra_pebble_ready(self, event) -> None:
```
gruyaume, line 141: Add typing

src/charm.py
```diff
@@ -116,19 +123,30 @@ def _get_database_relation_info(self) -> dict:
...
-            "endpoints": relation_data["endpoints"],
+            "username": relation_data.get("username"),
+            "password": relation_data.get("password"),
+            "endpoints": relation_data.get("endpoints"),
         }
 
-    def _update_layer(self) -> None:
-        """Updates the Pebble configuration layer and config if changed."""
-        self.unit.status = MaintenanceStatus("Applying pebble layer")
+    def _run_sql_migration(self, set_timeout) -> None:
```
gruyaume, line 131: Add typing

gruyaume (approved):
Minor comments about typing

nsklikas (approved):

tests/unit/test_charm.py
```diff
@@ -50,37 +53,62 @@ def setup_postgres_relation(harness):
...
+
     harness.charm.on.hydra_pebble_ready.emit(CONTAINER_NAME)
-    assert harness.charm.unit.status == WaitingStatus("Waiting for leadership")
+
+    assert (
+        "status_set",
+        "waiting",
+        "Unit waiting for leadership to run the migration",
+        {"is_app": False},
+    ) in harness._get_backend_calls()
```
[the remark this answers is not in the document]
natalian98: I'm only checking if that status was registered in calls rather than asserting the current status.

PR #27, opened 2023-02-06
Initial Kratos and UI integration

natalian98:
Initial integration with kratos and login ui charms:
- added hydra-kratos relation interface to provide `oauth2_provider_url` (hydra admin API endpoint)
- added configurable kratos_ui_url
- updated hydra config file
- updated unit tests
- updated readme according to the current state of integration

Integration testing will be covered in bundle tests.

TODO: Publish the library and update `LIBID` after an initial review.


src/hydra_endpoints_provider.py
```diff
@@ -0,0 +1,30 @@
...
+import logging
+
+from ops.framework import Object
+
+logger = logging.getLogger(__name__)
+
+RELATION_NAME = "endpoint-info"
+
+
+class HydraEndpointsProvider(Object):
```
gruyaume, line 15: Who will be the user/requirer of this interface?

lib/charms/hydra/v0/hydra_endpoints_info.py
```diff
@@ -0,0 +1,121 @@
...
+        super().__init__(self.message)
+
+
+class HydraEndpointsRequirer(Object):
+    def __init__(self, charm, relation_name: str = RELATION_NAME):
+        super().__init__(charm, relation_name)
+        self.charm = charm
+        self.relation_name = relation_name
+
+    def get_relation_data(self):
```
gruyaume, line 97: Add return typing

lib/charms/hydra/v0/hydra_endpoints_info.py
```diff
@@ -0,0 +1,121 @@
...
+from ops.model import Application
+
+# TODO: Update once the lib is published
+# # The unique Charmhub library identifier, never change it
+# LIBID = ""
+#
+# # Increment this major API version when introducing breaking changes
+# LIBAPI = 0
+#
+# # Increment this PATCH version before using `charmcraft publish-lib` or reset
```
gruyaume, line 66: Those will be auto-generated when you create the lib `charmcraft create-lib`, not the `publish-lib` meaning that it should not be done in a different PR.

config.yaml
```diff
@@ -0,0 +1,5 @@
+options:
```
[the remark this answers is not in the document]
natalian98: I agree, but it will be handled in kratos-ui-operator interface, since this url is needed by both kratos and hydra. So there will be another PR once implemented on kratos-ui side

lib/charms/hydra/v0/hydra_endpoints_info.py
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env python3
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
```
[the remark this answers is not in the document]
natalian98: I changed the name as suggested. As for the 2nd part of your comment, I first wanted to get a general opinion from the team. I am open to suggestions in `charm-relation-interfaces` once we're done with the initial integration

src/hydra_endpoints_provider.py
nsklikas, line 15: If we are going to use a library then this class should be placed in the same library as the `HydraEndpointsRequirer`

config.yaml
```diff
@@ -0,0 +1,5 @@
...
+  kratos_ui_url:
```
nsklikas, line 2: Perhaps `login_ui_url` would be a better name

README.md
```diff
@@ -85,8 +71,66 @@ IS LAST PAGE				true
...
+### PostgreSQL
+
 This charm requires a relation with [postgresql-k8s-operator](https://github.com/canonical/postgresql-k8s-operator).
 
+### Ingress
+
+The Hydra Operator offers integration with the [traefik-k8s-operator](https://github.com/canonical/traefik-k8s-operator) for ingress.
+Hydra has two APIs which can be exposed through ingress, the public API and the admin API.
+
+If you have a traefik deployed and configured in your hydra model, to provide ingress to the admin API run:
```
nsklikas, line 83:
```suggestion
If you have traefik deployed and configured in your hydra model, to provide ingress to the admin API run:
```

README.md
```diff
@@ -85,8 +71,66 @@ IS LAST PAGE				true
...
+juju deploy kratos-external-idp-integrator hydra-integrator --channel edge
+juju relate hydra-integrator kratos
+```
+
+Finally, provide the integrator charm with the client details:
+```console
+juju config hydra-integrator provider=generic client_id=<client-id> client_secret=<client-secret> issuer_url=http://{traefik_public_ip}/{model_name}-{hydra_app_name}
+```
+
+For example, if hydra is running in model named `test` behind a proxy with address `http://10.64.140.46:80/`, the `issuer_url` will be `http://10.64.140.46:80/test-hydra`.
```
nsklikas, line 130: This is no longer needed

src/charm.py
```diff
@@ -115,9 +134,12 @@ def _render_conf_file(self) -> None:
 
         rendered = template.render(
             db_info=self._get_database_relation_info(),
-            consent_url="http://localhost:4455/consent",
-            error_url="http://localhost:4455/error",
-            login_url="http://localhost:4455/login",
+            consent_url=join(self.config.get("login_ui_url"), "consent"),
+            error_url=join(self.config.get("login_ui_url"), "error"),
```
nsklikas, line 138: This would be 
```suggestion
            error_url=join(self.config.get("login_ui_url"), "oidc_error"),
```

src/charm.py
```diff
@@ -259,14 +317,21 @@ def _on_admin_ingress_ready(self, event: IngressPerAppReadyEvent) -> None:
...
             logger.info("This app's public ingress URL: %s", event.url)
 
+        self._update_config_restart_service(event)
+
     def _on_ingress_revoked(self, event: IngressPerAppRevokedEvent) -> None:
         if self.unit.is_leader():
             logger.info("This app no longer has ingress")
 
+        self._update_config_restart_service(event)
+        self._send_relation_info()
```
nsklikas, line 333: If something goes wrong in `_update_config_restart_service` (e.g. can't connect to the container or the db is not ready), the `_send_relation_info` method will still run and add the URLs in the databag.

This is not necessarily the wrong behavior, but we should be careful about it. 

Placing the URLs in the databag is meant to signal to Kratos that Hydra is ready or just that WHEN Hydra is up and ready that is where it will be found?

natalian98: I meant the second option you described: tell kratos that hydra will be available under this URL when ready. I also do it when there's no need to run `_update_config_restart_service` (for example `_on_hydra_relation_changed`). Hydra will then push its new endpoints whenever they get updated.
This is to avoid complexity - I think it's unnecessary to block kratos when the `hydra:kratos` relation is in place but hydra is not ready.
I also changed the default url to a predictable endpoint.

README.md
```diff
@@ -85,8 +71,48 @@ IS LAST PAGE				true
...
+
 This charm requires a relation with [postgresql-k8s-operator](https://github.com/canonical/postgresql-k8s-operator).
 
+### Ingress
+
+The Hydra Operator offers integration with the [traefik-k8s-operator](https://github.com/canonical/traefik-k8s-operator) for ingress.
+Hydra has two APIs which can be exposed through ingress, the public API and the admin API.
+
+If you have traefik deployed and configured in your hydra model, to provide ingress to the admin API run:
+```console
```
gruyaume, line 84: This somehow suggests that we need 2 traefik operators deployed. Is this accurate?

lib/charms/hydra/v0/hydra_endpoints.py
```diff
@@ -0,0 +1,139 @@
...
+logger = logging.getLogger(__name__)
+
+
+class HydraEndpointsProvider(Object):
+    """Provider side of the endpoint-info relation."""
+
+    def __init__(self, charm, relation_name=RELATION_NAME):
+        super().__init__(charm, relation_name)
+
+    def send_endpoint_relation_data(self, charm, admin_endpoint, public_endpoint):
```
gruyaume, line 80: Add typing

lib/charms/hydra/v0/hydra_endpoints.py
```diff
@@ -0,0 +1,139 @@
...
+        super().__init__(self.message)
+
+
+class HydraEndpointsRequirer(Object):
+    """Requirer side of the endpoint-info relation."""
+    def __init__(self, charm, relation_name: str = RELATION_NAME):
+        super().__init__(charm, relation_name)
+        self.charm = charm
+        self.relation_name = relation_name
+
```
gruyaume, line 114: You may want to include a custom event to be triggered when endpoints are available.

src/charm.py
```diff
@@ -67,7 +78,16 @@ def __init__(self, *args):
...
+        self.endpoints_provider = HydraEndpointsProvider(self)
+
         self.framework.observe(self.on.hydra_pebble_ready, self._on_hydra_pebble_ready)
+        self.framework.observe(self.on.config_changed, self._on_config_changed)
+        self.framework.observe(
+            self.on[RELATION_NAME].relation_created, self._on_hydra_relation_changed
+        )
+        self.framework.observe(
+            self.on[RELATION_NAME].relation_changed, self._on_hydra_relation_changed
+        )
```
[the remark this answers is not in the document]
gruyaume: Agreed

src/charm.py
```diff
@@ -142,6 +164,53 @@ def _run_sql_migration(self, set_timeout: bool) -> None:
...
+            self._container.start(self._container_name)
+            self.unit.status = ActiveStatus()
+            return
+
+        if self.model.relations[self._db_relation_name]:
+            self.unit.status = WaitingStatus("Waiting for database creation")
+        else:
+            self.unit.status = BlockedStatus("Missing required relation with postgresql")
+
+    def _send_relation_info(self) -> None:
```
gruyaume, line 195:
```suggestion
    def _update_hydra_endpoints_relation_data(self) -> None:
```

lib/charms/hydra/v0/hydra_endpoints.py
```diff
@@ -0,0 +1,139 @@
...
+    def get_relation_data(self) -> dict:
```
gruyaume, line 115:
```suggestion
    def get_hydra_endpoint(self) -> dict:
```

lib/charms/hydra/v0/hydra_endpoints.py
```diff
@@ -0,0 +1,139 @@
...
+        remote_app = [
+            app
+            for app in endpoints[0].data.keys()
+            if isinstance(app, Application) and not app._is_our_app
+        ][0]
+
+        data = endpoints[0].data[remote_app]
+
+        if not "admin_endpoint" in data:
+            logger.error("Missing admin endpoint in endpoint-info relation data")
```
gruyaume, line 131: Is this error needed knowing that we're raising right after with the same info

src/charm.py
```diff
@@ -142,6 +164,53 @@ def _run_sql_migration(self, set_timeout: bool) -> None:
...
+    def _update_config_restart_service(self, event: HookEvent) -> None:
+        if not self._container.can_connect():
+            event.defer()
+            logger.info("Cannot connect to Hydra container. Deferring the event.")
+            self.unit.status = WaitingStatus("Waiting to connect to Hydra container")
+            return
+
+        self.unit.status = MaintenanceStatus("Configuring resources")
+
+        try:
```
gruyaume, line 176: This try/except block can be in its own `_hydra_service_is_running` property/method:

```python
if not self._hydra_service_is_running:
    event.defer()
    return
```

src/charm.py
```diff
@@ -142,6 +164,53 @@ def _run_sql_migration(self, set_timeout: bool) -> None:
         stdout, _ = process.wait_output()
         logger.info(f"Executing automigration: {stdout}")
 
+    def _update_config_restart_service(self, event: HookEvent) -> None:
```
gruyaume, line 167: This method name is confusing. `update config restart service`? The method handles statuses and (sometimes) pushes a file to the workload.

src/charm.py
```diff
@@ -142,6 +164,53 @@ def _run_sql_migration(self, set_timeout: bool) -> None:
...
+
+        try:
+            self._container.get_service(self._container_name)
+        except (ModelError, RuntimeError):
+            event.defer()
+            self.unit.status = WaitingStatus("Waiting for Hydra service")
+            logger.info("Hydra service is absent. Deferring the event.")
+            return
+
+        if self.database.is_resource_created():
```
gruyaume, line 184: The method is inconsistent with somes `ifs` some `if nots` some `try/except` and ultimately hard to read. I think it'd look much cleaner with a consistent look:

```python
if not self._container.can_connect():
    event.defer()
    return
if not self._database_relation_is_created:
   self.unit.status = BlockedStatus("Missing required relation with postgresql")
   return
if not self._database_is_available:
   self.unit.status = WaitingStatus("Waiting for database to be available")
   event.defer()
   return
if not self._hydra_service_is_running:
    self.unit.status = WaitingStatus("Waiting for Hydra service")
    event.defer()
    return
if not <add any condition>:
   self.unit.status = WaitingStatus("bla bla bla")
   event.defer()
   return
self._container.push(self._hydra_config_path, self._render_conf_file(), make_dirs=True)
self._container.start(self._container_name)
self.unit.status = ActiveStatus()
```

tests/unit/test_charm.py
```diff
@@ -165,3 +242,47 @@ def test_ingress_relation_created(
         "port": port,
         "strip-prefix": "true",
     }
+
+
+def test_config_updated_on_ingress_relation_joined(
+    harness, mocked_kubernetes_service_patcher
+) -> None:
+    harness.begin()
```
gruyaume, line 250: I recommend to use `setUp` in order to avoid calling `begin()` and having to mock `kubernetes_service_patcher` in every unit test

src/charm.py
```diff
@@ -115,9 +134,12 @@ def _render_conf_file(self) -> None:
...
             db_info=self._get_database_relation_info(),
-            consent_url="http://localhost:4455/consent",
-            error_url="http://localhost:4455/error",
-            login_url="http://localhost:4455/login",
+            consent_url=join(self.config.get("login_ui_url"), "consent"),
+            error_url=join(self.config.get("login_ui_url"), "oidc_error"),
+            login_url=join(self.config.get("login_ui_url"), "login"),
+            hydra_public_url=self.public_ingress.url
+            if self.model.relations["public-ingress"]
+            else "http://127.0.0.1:4444/",
```
gruyaume, line 142:
```suggestion
            else f"http://127.0.0.1:{HYDRA_PUBLIC_PORT}/",
```

src/charm.py
```diff
@@ -67,7 +77,16 @@ def __init__(self, *args):
             strip_prefix=True,
         )
 
+        self.endpoints_provider = HydraEndpointsProvider(self)
+
         self.framework.observe(self.on.hydra_pebble_ready, self._on_hydra_pebble_ready)
+        self.framework.observe(self.on.config_changed, self._on_config_changed)
+        self.framework.observe(
+            self.on[RELATION_NAME].relation_created, self._on_hydra_relation_changed
```
gruyaume, line 85:
```suggestion
            self.on.endpoint_info_relation_created, self._on_hydra_relation_changed
```

gruyaume: I think it's useful to be specific in the relation name here aaand the variable `RELATION_NAME` is not great

lib/charms/hydra/v0/hydra_endpoints.py
```diff
@@ -0,0 +1,139 @@
...
+
+RELATION_NAME = "endpoint-info"
+INTERFACE_NAME = "hydra_endpoints"
+logger = logging.getLogger(__name__)
+
+
+class HydraEndpointsProvider(Object):
+    """Provider side of the endpoint-info relation."""
+
+    def __init__(self, charm, relation_name=RELATION_NAME):
```
gruyaume, line 77: Add typing

lib/charms/hydra/v0/hydra_endpoints.py
```diff
@@ -0,0 +1,139 @@
...
+This library provides a Python API for both requesting and providing public and admin endpoints.
+
+## Getting Started
+To get started using the library, you need to fetch the library using `charmcraft`.
+```shell
+cd some-charm
+charmcraft fetch-lib charms.hydra.v0.hydra_endpoints
+```
+
+To use the library from the provider side (Hydra):
```
gruyaume, line 16: I don't know if this bloc is particularly useful since the only provider will ever be this hydra charm. I'd focus on the requirer side

lib/charms/hydra/v0/hydra_endpoints.py
```diff
@@ -0,0 +1,139 @@
...
+In the `metadata.yaml` of the charm, add the following:
+```yaml
+provides:
+  endpoint-info:
+    interface: hydra_endpoints
+    description: Provides API endpoints to a related application
+```
+
+To use the library from the requirer side (Kratos):
```
gruyaume, line 25:
```suggestion
To use the library from the requirer side:
```

.github/workflows/tests.yaml
```diff
@@ -30,8 +30,9 @@ jobs:
...
         with:
           provider: microk8s
-          channel: 1.25/stable
-          bootstrap-options: '--agent-version=2.9.34'
+          channel: 1.25-strict/stable
+          juju-channel: 3.1
+          bootstrap-options: '--agent-version=3.1.0'
 
       - name: Run integration tests
         # set a predictable model name so it can be consumed by charm-logdump-action
```
gruyaume, line 38: You should add a job to automatically publish the lib once the charm has been published as well (probably in the `publish.yaml` file). Something like this should work:

```yaml
      - name: Publish libs
        env:
          CHARMCRAFT_AUTH: "${{ secrets.CHARMCRAFT_AUTH }}"
        run: |
          charmcraft publish-lib charms.hydra.v0.hydra_endpoints

```

src/charm.py
natalian98: Does `_handle_status_update_config` sound better?

src/charm.py
natalian98: Thanks for the suggestions, I updated it

natalian98:
Note: integration tests are possibly failing due to an intermittent issue with patching statefulset, see [charmhub channel](https://chat.charmhub.io/charmhub/pl/xhj3uoc6ztdbbpjfwn6r69ojta)

src/charm.py
```diff
@@ -142,6 +169,53 @@ def _run_sql_migration(self, set_timeout: bool) -> None:
...
         logger.info(f"Executing automigration: {stdout}")
 
+    def _handle_status_update_config(self, event: HookEvent) -> None:
+        if not self._container.can_connect():
+            event.defer()
+            logger.info("Cannot connect to Hydra container. Deferring the event.")
+            self.unit.status = WaitingStatus("Waiting to connect to Hydra container")
+            return
+
+        self.unit.status = MaintenanceStatus("Configuring resources")
```
gruyaume, line 179: This should probably set just before the `container.push` call, because in lots of cases we'll end up in waiting or blocked.

src/charm.py
```diff
@@ -142,6 +169,53 @@ def _run_sql_migration(self, set_timeout: bool) -> None:
...
+            self.unit.status = BlockedStatus("Missing required relation with postgresql")
+            return
+
+        if not self.database.is_resource_created():
+            self.unit.status = WaitingStatus("Waiting for database creation")
+            event.defer()
+            return
+
+        self._container.push(self._hydra_config_path, self._render_conf_file(), make_dirs=True)
+        self._container.start(self._container_name)
```
gruyaume, line 197: Shouldn't it be `restart` instead of `start`. If it is already running we what to restart right?

src/charm.py
```diff
@@ -142,6 +169,53 @@ def _run_sql_migration(self, set_timeout: bool) -> None:
...
+        self.unit.status = ActiveStatus()
+
+    def _update_hydra_endpoints_relation_data(self, event) -> None:
```
gruyaume, line 200: Add typing to `event`

src/charm.py
```diff
@@ -142,6 +169,53 @@ def _run_sql_migration(self, set_timeout: bool) -> None:
         stdout, _ = process.wait_output()
         logger.info(f"Executing automigration: {stdout}")
 
+    def _handle_status_update_config(self, event: HookEvent) -> None:
```
gruyaume, line 172: This method content looks much better now btw, thank you :)

src/charm.py
```diff
@@ -108,16 +123,28 @@ def _hydra_layer(self) -> Layer:
         }
         return Layer(layer_config)
 
+    @property
+    def _hydra_service_is_running(self) -> bool:
+        try:
+            self._container.get_service(self._container_name)
```
gruyaume, line 129: `get_service` won't tell you that the service is running but that it exists. you will probably want to have:
`service = self._container.get_service(self._container_name)` and instead of returning `True`, returning `service.is_running()`

src/charm.py
```diff
@@ -142,6 +169,53 @@ def _run_sql_migration(self, set_timeout: bool) -> None:
...
+        )
+        self.endpoints_provider.send_endpoint_relation_data(
+            self.app, admin_endpoint[0], public_endpoint[0]
+        )
+
+        logger.info(
+            f"Sent endpoints info: public - {public_endpoint[0]} admin - {admin_endpoint[0]}"
+        )
+
     def _on_hydra_pebble_ready(self, event: WorkloadEvent) -> None:
```
gruyaume, line 219: Is there a reason why pebble ready calls this same event handler `_handle_status_update_config` it seems to be doing the exact same thing right?

gruyaume: You could simply add the pebble layer part to `_handle_status_update_config` ex.
```
if the existing pebble layer is not the same as `pebble_layer`:
    add pebble layer
restart service
```

README.md
nsklikas: Just to be clear, we don't _need_ it but the current traefik/ingress relation interface does not allow us to have expose 2 ports exposed from the same traefik instance. 

Also you don't need to expose the admin API to have a working hydra installation.

src/charm.py
natalian98: With the latest updates hydra adds a layer and replans the container while in this status, so I think it's fine to keep it that way?

src/charm.py
natalian98: It made sense to `start` because before my latest changes some other event could be run before pebble started the service, resulting in pebble error `service "hydra" has never been started` while trying to restart. Now it's ok to restart in `_handle_status_update_config` but I need to keep `start` `_on_database_created`.

src/charm.py
natalian98: Wrong naming, I just want to check if the service is present, not that it's running. Renamed to `_hydra_service_is_created`

lib/charms/hydra/v0/hydra_endpoints.py
```diff
@@ -0,0 +1,169 @@
...
+    def _on_provider_endpoint_relation_created(self, event: RelationCreatedEvent):
+        self.on.ready.emit()
+
+    def _on_provider_endpoint_relation_changed(self, event: RelationChangedEvent):
+        self.on.ready.emit()
+
+    def send_endpoint_relation_data(
+        self, charm, admin_endpoint: str, public_endpoint: str
+    ) -> None:
+        """Updates relation with endpoints info."""
```
nsklikas, line 111: Shouldn't this check if the unit is leader?

src/charm.py
```diff
@@ -152,26 +179,63 @@ def _on_hydra_pebble_ready(self, event: WorkloadEvent) -> None:
...
-        else:
-            self.unit.status = BlockedStatus("Missing required relation with postgresql")
+            event.defer()
+            return
+
+        self._container.push(self._hydra_config_path, self._render_conf_file(), make_dirs=True)
+        self._container.restart(self._container_name)
+        self.unit.status = ActiveStatus()
+
+    def _update_hydra_endpoints_relation_data(self, event: RelationEvent) -> None:
```
nsklikas, line 213: Not sure if we actually care about the types(probably not?), but the correct type for the event here is `HookEvent`

natalian98: Are you sure? This method can be called by ingress or hydra endpoints relation events, which are all of `RelationEvent` class.
`RelationEvent` is based on `HookEvent` though

lib/charms/hydra/v0/hydra_endpoints.py
natalian98: Yeah I think it makes sense to check it. Thanks

README.md
jnsgruk: Hmm. This is a bit odd - @PietroPasotti do you have any insight here?

Feels like we could deploy with 1x `traefik-k8s` and 2x `traefik-route-k8s` (one for each endpoint)?

src/charm.py
```diff
@@ -108,16 +124,28 @@ def _hydra_layer(self) -> Layer:
...
             db_info=self._get_database_relation_info(),
-            consent_url="http://localhost:4455/consent",
-            error_url="http://localhost:4455/error",
-            login_url="http://localhost:4455/login",
+            consent_url=join(self.config.get("login_ui_url"), "consent"),
+            error_url=join(self.config.get("login_ui_url"), "oidc_error"),
+            login_url=join(self.config.get("login_ui_url"), "login"),
+            hydra_public_url=self.public_ingress.url
+            if self.model.relations["public-ingress"]
+            else f"http://127.0.0.1:{HYDRA_PUBLIC_PORT}/",
```
nsklikas, line 148: I believe this should be configurable, someone might want to provide non-charmed ingress or something like that.

Most charms provide a configuration option like `external_url`, do we want to allow that?

natalian98: I'm fine with making this configurable, but in that case we'd also need to provide the same possibility for the admin endpoint and I'm not sure we want to add that many config options when we decided to recommend using charmed traefik. @massimilianogo any thoughts?

README.md
PietroPasotti: Ummm I think you should be able to configure a single traefik-route instance to expose multiple ports, or indeed, you'd need two traefik instances, one for each port/endpoint.

src/charm.py
massimilianogo: Technically that should be true for all of the charms we have...and that is a lot of config.
In general I think that an admin will use a single ingress for the entire deployment - either our Traefik or something else.
My suggestion is to only have traefik for now and in parallel open a feature request (I can do that) in the relevant board to track the feature. If there is a need we can add it to the backlog and prioritize it accordingly.

README.md
massimilianogo: I could be wrong but I believe that the initial decision of using 2 x traefik-route-k8s was taken after a discussion about the traefik-k8s limitations with Dominik and @simskij 
Happy to update our approach if you can point us to a working implementation or provide more guidance

nsklikas (dismissed):
LGTM, I only have a small comment. Change it only if it makes sense.

lib/charms/hydra/v0/hydra_endpoints.py
```diff
@@ -0,0 +1,172 @@
...
+
+        events = self._charm.on[relation_name]
+        self.framework.observe(events.relation_joined, self._on_provider_endpoint_relation_joined)
+        self.framework.observe(events.relation_created, self._on_provider_endpoint_relation_joined)
+        self.framework.observe(
+            events.relation_changed, self._on_provider_endpoint_relation_changed
+        )
+
+    def _on_provider_endpoint_relation_joined(self, event: RelationJoinedEvent):
+        self.on.ready.emit()
```
gruyaume, line 100: Is this event going to be used anywhere in the charm? If not I'd simply get rid of it

lib/charms/hydra/v0/hydra_endpoints.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def _on_provider_endpoint_relation_created(self, event: RelationCreatedEvent):
+        self.on.ready.emit()
+
+    def _on_provider_endpoint_relation_changed(self, event: RelationChangedEvent):
+        self.on.ready.emit()
+
+    def send_endpoint_relation_data(
+        self, charm, admin_endpoint: str, public_endpoint: str
```
gruyaume, line 109: Add typing for `charm`

nsklikas (approved):

PR #29, opened 2023-03-03
Add oauth library

nsklikas:
Adds the library implementing https://github.com/canonical/charm-relation-interfaces/blob/main/interfaces/oauth/v0/README.md

TODO:
- [x] Add ca_chain to relation (Will be done in future PR)
- [x] More tests?
- [x] Do we want integration tests (Will be done in future PR)

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -46,12 +46,16 @@ jobs:
         run: juju status --relations
         if: failure()
 
-      - name: Dump logs
-        uses: canonical/charm-logdump-action@main
+      - name: Get juju status
+        run: juju debug-log --replay --include hydra/0
         if: failure()
-        with:
-          app: hydra
-          model: testing
+
+      # - name: Dump logs
+      #   uses: canonical/charm-logdump-action@main
+      #   if: failure()
+      #   with:
+      #     app: hydra
+      #     model: testing
 
       - name: Get hydra container logs
         run: kubectl logs hydra-0 -c hydra -n testing
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -50,13 +50,6 @@ jobs:
         run: juju debug-log --replay --include hydra/0
         if: failure()
 
-      # - name: Dump logs
-      #   uses: canonical/charm-logdump-action@main
-      #   if: failure()
-      #   with:
-      #     app: hydra
-      #     model: testing
-
       - name: Get hydra container logs
         run: kubectl logs hydra-0 -c hydra -n testing
 
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -46,7 +46,7 @@ jobs:
         run: juju status --relations
         if: failure()
 
-      - name: Get juju status
+      - name: Get juju logs
         run: juju debug-log --replay --include hydra/0
         if: failure()
 
--- a/lib/charms/hydra/v0/oauth.py
+++ b/lib/charms/hydra/v0/oauth.py
@@ -612,10 +612,8 @@ class OAuthProvider(Object):
         return f"client_secret_{relation.id}"
 
     def _on_relation_departed(self, event: RelationDepartedEvent) -> None:
-        # The relation has been departed, delete the juju secret
         secret = self.model.get_secret(label=self._get_secret_label(event.relation))
         secret.remove_all_revisions()
-        # The provider needs to delete the client
         self.on.client_deleted.emit(event.relation.id)
 
     def _create_juju_secret(self, client_secret: str, relation: Relation) -> Secret:
--- a/lib/charms/hydra/v0/oauth.py
+++ b/lib/charms/hydra/v0/oauth.py
@@ -1,0 +1,545 @@
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""# Oauth Library.
+
+This library is designed to enable applications to register OAuth2/OIDC
+clients with an OIDC Provider through the `oauth` interface.
+
+## Getting started
+
+To get started using this library you just need to fetch the library using `charmcraft`. **Note
+that you also need to add `jsonschema` to your charm's `requirements.txt`.**
+
+```shell
+cd some-charm
+charmcraft fetch-lib charms.hydra.v0.oauth
+EOF
+```
+
+Then, to initialize the library:
+```python
+# ...
+from charms.hydra.v0.kubernetes_service_patch import ClientConfig, OAuthRequirer
+
+OAUTH = "oauth"
+OAUTH_SCOPES = "openid email"
+OAUTH_GRANT_TYPES = ["authorization_code"]
+
+class SomeCharm(CharmBase):
+  def __init__(self, *args):
+    # ...
+    self.oauth = OAuthRequirer(self, client_config, relation_name=OAUTH)
+
+    self.framework.observe(self.oauth.on.client_credentials_changed, self._configure_application)
+    self.framework.observe(self.oauth.on.provider_config_changed, self._configure_application)
+    # ...
+
+    def _on_ingress_ready(self, event):
+        self.external_url = "https://example.com
+        self._set_client_config()
+
+    def _set_client_config(self):
+        client_config = ClientConfig(
+            join(self.external_url, "/oauth/callback"),
+            OAUTH_SCOPES,
+            OAUTH_GRANT_TYPES,
+        )
+        self.oauth.update_client_config(client_config)
+```
+"""
+
+import json
+import logging
+import re
+from dataclasses import asdict, dataclass, field
+
+import jsonschema
+from ops.framework import EventBase, EventSource, Object, ObjectEvents
+
+# The unique Charmhub library identifier, never change it
+LIBID = "a3a301e325e34aac80a2d633ef61fe97"
+
+# Increment this major API version when introducing breaking changes
+LIBAPI = 0
+
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 1
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_RELATION_NAME = "oauth"
+ALLOWED_GRANT_TYPES = ["authorization_code", "refresh_token", "client_credentials"]
+ALLOWED_CLIENT_AUTHN_METHODS = ["client_secret_basic", "client_secret_post"]
+CLIENT_SECRET_FIELD = "secret"
+
+url_regex = re.compile(
+    r"^https://"  # https://
+    r"(?:(?:[A-Z0-9](?:[A-Z0-9-]{0,61}[A-Z0-9])?\.)+(?:[A-Z]{2,6}\.?|"
+    r"[A-Z0-9-]{2,}\.?)|"  # domain...
+    r"\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})"  # ...or ip
+    r"(?::\d+)?"  # optional port
+    r"(?:/?|[/?]\S+)$",
+    re.IGNORECASE,
+)
+
+OAUTH_PROVIDER_JSON_SCHEMA = {
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "$id": "https://canonical.github.io/charm-relation-interfaces/interfaces/oauth/schemas/provider.json",
+    "type": "object",
+    "properties": {
+        "issuer_url": {
+            "type": "string",
+        },
+        "authorization_endpoint": {
+            "type": "string",
+        },
+        "token_endpoint": {
+            "type": "string",
+        },
+        "introspection_endpoint": {
+            "type": "string",
+        },
+        "userinfo_endpoint": {
+            "type": "string",
+        },
+        "jwks_endpoint": {
+            "type": "string",
+        },
+        "scope": {
+            "type": "string",
+        },
+        "client_id": {
+            "type": "string",
+        },
+        "client_secret_id": {
+            "type": "string",
+        },
+        "groups": {"type": "string", "default": None},
+        "ca_chain": {"type": "array", "items": {"type": "string"}, "default": []},
+    },
+    "required": [
+        "issuer_url",
+        "authorization_endpoint",
+        "token_endpoint",
+        "introspection_endpoint",
+        "userinfo_endpoint",
+        "jwks_endpoint",
+        "scope",
+    ],
+}
+OAUTH_REQUIRER_JSON_SCHEMA = {
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "$id": "https://canonical.github.io/charm-relation-interfaces/interfaces/oauth/schemas/requirer.json",
+    "type": "object",
+    "properties": {
+        "redirect_uri": {
+            "type": "string",
+            "default": None,
+        },
+        "audience": {"type": "array", "default": [], "items": {"type": "string"}},
+        "scope": {"type": "string", "default": None},
+        "grant_types": {
+            "type": "array",
+            "default": None,
+            "items": {
+                "enum": ["authorization_code", "client_credentials", "refresh_token"],
+                "type": "string",
+            },
+        },
+        "token_endpoint_auth_method": {
+            "type": "string",
+            "enum": ["client_secret_basic", "client_secret_post"],
+            "default": "client_secret_basic",
+        },
+    },
+    "required": ["redirect_uri", "audience", "scope", "grant_types", "token_endpoint_auth_method"],
+}
+
+
+class ClientConfigError(Exception):
+    """Emitted when invalid client config is provided."""
+
+    pass
+
+
+class DataValidationError(RuntimeError):
+    """Raised when data validation fails on relation data."""
+
+
+def _load_data(data, schema=None):
+    """Parses nested fields and checks whether `data` matches `schema`."""
+    ret = {}
+    for k, v in data.items():
+        try:
+            ret[k] = json.loads(v)
+        except json.JSONDecodeError as e:
+            ret[k] = v
+
+    if schema:
+        _validate_data(ret, schema)
+    return ret
+
+
+def _dump_data(data, schema=None):
+    if schema:
+        _validate_data(data, schema)
+
+    ret = {}
+    for k, v in data.items():
+        if isinstance(v, (list, dict)):
+            try:
+                ret[k] = json.dumps(v)
+            except json.JSONDecodeError as e:
+                raise DataValidationError(f"Failed to encode relation json: {e}")
+        else:
+            ret[k] = v
+    return ret
+
+
+def _validate_data(data, schema):
+    """Checks whether `data` matches `schema`.
+
+    Will raise DataValidationError if the data is not valid, else return None.
+    """
+    try:
+        jsonschema.validate(instance=data, schema=schema)
+    except jsonschema.ValidationError as e:
+        raise DataValidationError(data, schema) from e
+
+
+@dataclass
+class ClientConfig:
+    """Helper class containing a client's configuration."""
+
+    redirect_uri: str
+    scope: str
+    grant_types: list[str]
+    audience: list[str] = field(default_factory=lambda: [])
+    token_endpoint_auth_method: str = "client_secret_basic"
+
+    def validate(self):
+        """Validate the client configuration."""
+        # Validate redirect_uri
+        if not re.match(url_regex, self.redirect_uri):
+            raise ClientConfigError(f"Invalid URL {self.redirect_uri}")
+
+        # Validate grant_types
+        for grant_type in self.grant_types:
+            if grant_type not in ALLOWED_GRANT_TYPES:
+                raise ClientConfigError(
+                    f"Invalid grant_type {grant_type}, must be one " f"of {ALLOWED_GRANT_TYPES}"
+                )
+
+        # Validate client authentication methods
+        if self.token_endpoint_auth_method not in ALLOWED_CLIENT_AUTHN_METHODS:
+            raise ClientConfigError(
+                f"Invalid client auth method {self.token_endpoint_auth_method}, "
+                f"must be one of {ALLOWED_CLIENT_AUTHN_METHODS}"
+            )
+
+
+class ClientCredentialsChangedEvent(EventBase):
+    """Event to notify the charm that the client credentials changed."""
+
+    def __init__(self, handle, client_id, client_secret_id):
+        super().__init__(handle)
+        self.client_id = client_id
+        self.client_secret_id = client_secret_id
+
+    def snapshot(self):
+        """Save event."""
+        return {
+            "client_id": self.client_id,
+            "client_secret_id": self.client_secret_id,
+        }
+
+    def restore(self, snapshot):
+        """Restore event."""
+        self.client_id = snapshot["client_id"]
+        self.client_secret_id = snapshot["client_secret_id"]
+
+
+class ProviderConfigChangedEvent(EventBase):
+    """Event to notify the charm that the provider's configuration changed."""
+
+
+class OAuthRequirerEvents(ObjectEvents):
+    """Event descriptor for events raised by `OAuthRequirerEvents`."""
+
+    client_credentials_changed = EventSource(ClientCredentialsChangedEvent)
+    provider_config_changed = EventSource(ProviderConfigChangedEvent)
+
+
+class OAuthRequirer(Object):
+    """Register an oauth client."""
+
+    on = OAuthRequirerEvents()
+
+    def __init__(self, charm, client_config=None, relation_name=DEFAULT_RELATION_NAME):
+        super().__init__(charm, relation_name)
+        self._charm = charm
+        self._relation_name = relation_name
+        self._client_config = client_config
+        events = self._charm.on[relation_name]
+        self.framework.observe(events.relation_created, self._on_relation_created_event)
+        self.framework.observe(events.relation_changed, self._on_relation_changed_event)
+
+    def _on_relation_created_event(self, event):
+        try:
+            self._update_relation_data(self._client_config, event.relation.id)
+        except Exception:
+            pass
+
+    def _on_relation_changed_event(self, event):
+        if not self.model.unit.is_leader():
+            return
+
+        data = _load_data(event.relation.data[event.app], OAUTH_PROVIDER_JSON_SCHEMA)
+
+        client_id = data.get("client_id")
+        client_secret_id = data.get("client_secret_id")
+        if not client_id or not client_secret_id:
+            # This probably means that the Provider just set its endpoint,
+            # but it could also mean that the providers removed the client credentials
+            # from the databag. We could do something similar to
+            # data_platform_libs/v0/data_interfaces:diff if we wanted to
+            # TODO
+            self.on.provider_config_changed.emit()
+            return
+
+        if client_secret_id:
+            self.on.client_credentials_changed.emit(client_id, client_secret_id)
+        else:
+            # TODO: log some error?
+            pass
+
+    def _update_relation_data(self, client_config, relation_id):
+        if not self.model.unit.is_leader():
+            return
+
+        try:
+            client_config.validate()
+        except ClientConfigError as e:
+            # emit error event
+            logger.info(e)
+            return
+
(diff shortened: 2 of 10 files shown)


lib/charms/hydra/v0/oauth.py
```diff
@@ -168,21 +171,21 @@ class DataValidationError(RuntimeError):
     """Raised when data validation fails on relation data."""
```
gruyaume, line 172: You'll need to bump the `LIBPATCH` version

lib/charms/hydra/v0/oauth.py
```diff
@@ -168,21 +171,21 @@ class DataValidationError(RuntimeError):
...
 
 
-def _load_data(data, schema=None):
+def _load_data(data: Dict, schema: Optional[Dict] = None):
```
gruyaume, line 174: Add return typing

lib/charms/hydra/v0/oauth.py
```diff
@@ -168,21 +171,21 @@ class DataValidationError(RuntimeError):
...
+        except json.JSONDecodeError:
             ret[k] = v
 
     if schema:
         _validate_data(ret, schema)
     return ret
 
 
-def _dump_data(data, schema=None):
+def _dump_data(data: Dict, schema: Optional[Dict] = None):
```
gruyaume, line 188: Add return typing

src/charm.py
```diff
@@ -340,8 +348,12 @@ def _create_client(self, client_config, metadata=None):
 
         logger.info(cmd)
 
-        process = self._container.exec(cmd)
-        stdout, _ = process.wait_output()
+        try:
+            process = self._container.exec(cmd)
+            stdout, _ = process.wait_output()
+        except ExecError as err:
```
gruyaume, line 354: I'm not 100% sure but I think the `ExecError` would only be raised on the `wait_output` call so I'd remove the `exec` from the try/except block

lib/charms/hydra/v0/oauth.py
nsklikas: Why? This is the first version of the library

src/charm.py
nsklikas: You are right, `exec` can only throw errors when the operator fails to connect to the pebble websocket. Is it ok now or do should we split the try/except?

lib/charms/hydra/v0/oauth.py
```diff
@@ -0,0 +1,569 @@
...
+
+class ProviderConfigChangedEvent(EventBase):
+    """Event to notify the charm that the provider's configuration changed."""
+
+
+class OAuthRequirerEvents(ObjectEvents):
+    """Event descriptor for events raised by `OAuthRequirerEvents`."""
+
+    client_credentials_changed = EventSource(ClientCredentialsChangedEvent)
+    provider_config_changed = EventSource(ProviderConfigChangedEvent)
```
nsklikas, line 275: @gruyaume 
There are two sets of information in the providers databag, the provider's endpoints and the client's credentials (if a client has been registered). For most client applications it should make no difference and they should want to be notified only when the full set of information is available (provider endpoints + client credentials).

Semantically it makes sense to have two separate events, as the provider will always first place the provider_config in the databag and then the client credentials.

Do you think it makes sense to have 2 events or should we only emit one when all the information is available?

lib/charms/hydra/v0/oauth.py
gruyaume: My bad! Indeed

src/charm.py
gruyaume: I'd put only the `wait_output` inside of the try/except block and place the `exec` right before

src/charm.py
```diff
@@ -259,14 +274,109 @@ def _on_admin_ingress_ready(self, event: IngressPerAppReadyEvent) -> None:
...
+        self.oauth.set_client_credentials_in_relation_data(
+            event.relation_id, client["client_id"], client["client_secret"]
+        )
+
+    def _on_client_config_changed(self, event: ClientConfigChangedEvent) -> None:
+        ...
+        # client_config = event.to_client_config()
+        # self._create_client(client_config, metadata=f"{{\"relation_id\": {event.relation_id}}}")
+
+    def _create_client(
```
gruyaume, line 316: We may want a different class/file for the `hydra` specific calls and have this content in the init. If there are other calls that require using the hydra client they can all be there.

lib/charms/hydra/v0/oauth.py
gruyaume: I'd say place yourself in the charm author point of view, if there's nothing you can do with each event individually and when receiving them you need to create logic around checking whether the *other* info is also set, I'd say have only 1 event when all the info is necessary. If there's actions you can take with each of the event individually, sure let's use the 2 events.

lib/charms/hydra/v0/oauth.py
```diff
@@ -0,0 +1,610 @@
...
+    def _on_relation_changed_event(self, event: RelationChangedEvent) -> None:
+        if not self.model.unit.is_leader():
+            return
+
+        data = _load_data(event.relation.data[event.app], OAUTH_PROVIDER_JSON_SCHEMA)
+
+        client_id = data.get("client_id")
+        client_secret_id = data.get("client_secret_id")
+        if not client_id or not client_secret_id:
+            # The client credentials are not ready yet, so we do nothing
```
gruyaume, line 328: Do you want to add a log here?

lib/charms/hydra/v0/oauth.py
```diff
@@ -0,0 +1,610 @@
...
+        data.pop("client_secret_id", None)
+        return data
+
+    def get_client_secret(self, client_secret_id: str) -> Secret:
+        """Get the client_secret."""
+        client_secret = self.model.get_secret(id=client_secret_id)
+        return client_secret
+
+    def update_client_config(
+        self, client_config: ClientConfig, relation_id: Optional[int] = None
```
gruyaume, line 378: In which cases will the user of this method know the relation id?

gruyaume: Also, the `client_config` was provided in the init, why re-set it here?

tests/unit/test_oauth_requirer.py
```diff
@@ -0,0 +1,176 @@
...
+
+def test_data_in_relation_bag_on_joined(harness):
+    relation_id = harness.add_relation("oauth", "provider")
+
+    relation_data = harness.get_relation_data(relation_id, harness.model.app.name)
+
+    assert _load_data(relation_data) == CLIENT_CONFIG
+
+
+def test_oauth_info_changed_emitted_on_client_creation(harness, provider_info):
```
gruyaume, line 78: This unit test is very long. Ideally, the goal is for a test to only have 1 reason to crash. While it's fine to have more than 1 `assert` (ex. if there are multiple values you want to check in a dict), you really only want to validate 1 behavior.

tests/unit/test_charm.py
```diff
@@ -261,3 +273,178 @@ def test_config_updated_on_ingress_relation_joined(harness) -> None:
...
+
+def test_provider_info_in_databag_when_ingress_then_oauth_relation(harness):
+    harness.begin()
+    harness.set_can_connect(CONTAINER_NAME, True)
+
+    setup_ingress_relation(harness, "public")
+    setup_ingress_relation(harness, "admin")
+    relation_id, _ = setup_oauth_relation(harness)
+
+    app_data = harness.get_relation_data(relation_id, harness.charm.app)
```
gruyaume, line 286: You may want to separate concerns between the charm code and the lib code. Asserting content of relation data is probably preferable to test in the lib tests since it's the lib that does this. In the charm tests you may want to simply validate that the correct lib method is called.

lib/charms/hydra/v0/oauth.py
```diff
@@ -0,0 +1,610 @@
...
+        self.on.client_deleted.emit(event.relation.id)
+
+    def _create_juju_secret(self, client_secret: str, relation: Relation) -> Secret:
+        """Create a juju secret and grant it to a relation."""
+        secret = {CLIENT_SECRET_FIELD: client_secret}
+        juju_secret = self.model.app.add_secret(secret, label="client_secret")
+        juju_secret.grant(relation)
+        return juju_secret
+
+    def set_provider_info_in_relation_data(self, data: Dict) -> None:
```
gruyaume, line 588: Here I'd be more restrictive in terms of what the method allows as an input, you can list all the parameters that are expected.

nsklikas:
There are some features missing from the current implementation, but I think it's best to add them in separate PRs in order not to overload this one:
- Currently traefik does not pass its ca_chain from the relation interface, making it impossible to get the ca_chain. We will add that feature in the future.
- When the relation is broken, the client should be deleted. This is not handled right now, as when we get the relation broken event we have no way of knowing which client we must delete. I think the most common solution to such a problem would be to use the peer relation to store a mapping from `relation_id` to `client_id` (@gruyaume any thoughts on this?).

lib/charms/hydra/v0/oauth.py
nsklikas: > In which cases will the user of this method know the relation id?

The user(charm) may know the `relation_id` from an event. Tbh I don't really like having the relation_id as an argument like this, but the juju/ops framework primitives around getting relations are a little weird and as far as I can tell there is no clean way of enforcing that only on relation will be created. Having more than one relations from a single oauth requirer relation does not really make sense. 

I would be OK with making the assumption in the code that there will always be at most 1 relation available (thus we would not need the relation_id to get the relation) and just having a warning in the docs.

> Also, the `client_config` was provided in the init, why re-set it here?

One of the client configuration needed is the `redirect_uri`. IMO most of the time the requirer will not know that value when initializing the `OAuthRequirer` class as that would be provided to the requirer later either from the ingress relation or user configuration. This means that the charm will need to call this function in order to update the client_config it provided.

tests/unit/test_charm.py
nsklikas: But you also need to be sure that the charm is passing the correct information to the library. I could mock the library methods and assert that they are called with the correct arguments, but it seems simpler to validate that the final state is correct (ie the correct information is placed in the databag).

massimilianogo:
> Currently traefik does not pass its ca_chain from the relation interface, making it impossible to get the ca_chain. We will add that feature in the future.

I agree, let's keep this as a separate PR

tests/unit/test_oauth_requirer.py
nsklikas: I split it into 2 tests. 
Currently, I am checking that the correct event is emitted, but also that the `client_secret_id` in the event corresponds to the correct juju secret, is that ok with you or do you think I should try to break it down even further?

tests/unit/test_charm.py
gruyaume: Exact, I think you should do the first thing you are proposing (mocking the lib and asserting you are calling it with the correct args). If you are validating for the relation data you would be duplicating tests as you'd also be testing this in the lib tests.

gruyaume: Another way to see this. If the lib doesn't write the relation data correctly, you want only 1 test to crash, ideally a test in the lib, not on something that is dependent on the lib.

src/charm.py
```diff
@@ -302,20 +327,180 @@ def _on_admin_ingress_ready(self, event: IngressPerAppReadyEvent) -> None:
...
+        """Update an oauth2 client."""
+        cmd = self._client_cmd_prefix("update") + self._client_config_to_cmd(
+            client_config, metadata
+        )
+
+        stdout, _ = self._run_cmd(cmd)
+        logger.info(f"Successfully updated client: {client_config.client_id}")
+        return json.loads(stdout)
+
+    def delete_client(
```
gruyaume, line 490: All other client related methods focus on 1 client, I'd do the same here

gruyaume: In the same vein, update_client is just for 1 client.

src/charm.py
```diff
@@ -302,20 +327,180 @@ def _on_admin_ingress_ready(self, event: IngressPerAppReadyEvent) -> None:
...
+            flags.append("--metadata")
+            if isinstance(metadata, dict):
+                flags.append(json.dumps(metadata))
+            elif isinstance(metadata, str):
+                flags.append(metadata)
+            else:
+                raise ValueError("Invalid metadata")
+        return flags
+
+    def _client_cmd_prefix(self, action: str):
```
gruyaume, line 455: Add return typing

src/charm.py
```diff
@@ -302,20 +327,180 @@ def _on_admin_ingress_ready(self, event: IngressPerAppReadyEvent) -> None:
...
+        self, client_ids: List[str], metadata: Optional[Union[Dict, str]] = None
+    ) -> Optional[Dict]:
+        """Delete one or more oauth2 client."""
+        cmd = self._client_cmd_prefix("delete") + client_ids
+
+        stdout, _ = self._run_cmd(cmd)
+        logger.info(f"Successfully deleted clients: {stdout}")
+        return json.loads(stdout)
+
+    def _run_cmd(self, cmd, timeout=20):
```
gruyaume, line 500: Add return typing

nsklikas:
On the requirer side do we want to emit an event once the relation is broken or are the charms expected to handle that on their own?

.github/workflows/tests.yaml
```diff
@@ -46,12 +46,16 @@ jobs:
         run: juju status --relations
         if: failure()
 
-      - name: Dump logs
-        uses: canonical/charm-logdump-action@main
+      - name: Get juju status
```
natalian98, line 49:
```suggestion
      - name: Get juju logs
```

src/charm.py
```diff
@@ -302,20 +327,183 @@ def _on_admin_ingress_ready(self, event: IngressPerAppReadyEvent) -> None:
...
+    def _on_oauth_relation_created(self, event: RelationCreatedEvent) -> None:
+        self._update_endpoint_info()
+
+    def _on_client_created(self, event: ClientCreatedEvent) -> None:
+        if not self._container.can_connect():
+            event.defer()
+            return
+
+        try:
+            self._container.get_service(self._container_name)
```
natalian98, line 357: There is a similar method `_hydra_service_is_created`, you could reuse it
```python
if not self._hydra_service_is_created:
  event.defer()
  return
```

src/charm.py
```diff
@@ -302,20 +327,183 @@ def _on_admin_ingress_ready(self, event: IngressPerAppReadyEvent) -> None:
...
+            event.relation_id, client["client_id"], client["client_secret"]
+        )
+
+    def _on_client_changed(self, event: ClientChangedEvent) -> None:
+        if not self._container.can_connect():
+            event.defer()
+            return
+
+        try:
+            self._container.get_service(self._container_name)
```
natalian98, line 386: same here:
```python
if not self._hydra_service_is_created:
  event.defer()
  return
```

src/charm.py
```diff
@@ -302,20 +327,183 @@ def _on_admin_ingress_ready(self, event: IngressPerAppReadyEvent) -> None:
...
+    def create_client(
+        self, client_config: ClientConfig, metadata: Optional[Union[Dict, str]] = None
+    ) -> Optional[Dict]:
+        """Create an oauth2 client."""
+        cmd = self._client_cmd_prefix("create") + self._client_config_to_cmd(
+            client_config, metadata
+        )
+
+        stdout, _ = self._run_cmd(cmd)
+        logger.info("Successfully created client")
```
natalian98, line 475: Do we want to log the created client id as well?

lib/charms/hydra/v0/oauth.py
```diff
@@ -0,0 +1,671 @@
+# Copyright 2021 Canonical Ltd.
```
natalian98, line 1:
```suggestion
# Copyright 2023 Canonical Ltd.
```

lib/charms/hydra/v0/oauth.py
```diff
@@ -0,0 +1,671 @@
...
+class SomeCharm(CharmBase):
+  def __init__(self, *args):
+    # ...
+    self.oauth = OAuthRequirer(self, client_config, relation_name=OAUTH)
+
+    self.framework.observe(self.oauth.on.oauth_info_changed, self._configure_application)
+    # ...
+
+    def _on_ingress_ready(self, event):
+        self.external_url = "https://example.com
```
natalian98, line 38:
```suggestion
        self.external_url = "https://example.com"
```

lib/charms/hydra/v0/oauth.py
```diff
@@ -0,0 +1,671 @@
...
+            self._update_relation_data(self._client_config, event.relation.id)
+        except ClientConfigError as e:
+            self.on.invalid_client_config.emit(e.args[0])
+
+    def _on_relation_changed_event(self, event: RelationChangedEvent) -> None:
+        if not self.model.unit.is_leader():
+            return
+
+        data = event.relation.data[event.app]
+        if not data:
```
natalian98, line 324: Maybe add a log entry that the relation data is not available?

gruyaume:
> There are some features missing from the current implementation, but I think it's best to add them in separate PRs in order not to overload this one:
> 
> * Currently traefik does not pass its ca_chain from the relation interface, making it impossible to get the ca_chain. We will add that feature in the future.
> * When the relation is broken, the client should be deleted. This is not handled right now, as when we get the relation broken event we have no way of knowing which client we must delete. I think the most common solution to such a problem would be to use the peer relation to store a mapping from `relation_id` to `client_id` (@gruyaume any thoughts on this?).

- It's fine not to add "everything" from a feature in a PR, as long as the PR doesn't break the current behaviors. Make sure you have a user story that covers the changes that weren't included here. Either your current user story or a new one.
- The question of the relation broken is a good one. I'm not a fan of having to maintain a mapping like this one and would avoid it if possible. In `hydra` when we create a client, it's not possible to add a tag right? That could be useful to know which one to delete.

nsklikas:
>     * The question of the relation broken is a good one. I'm not a fan of having to maintain a mapping like this one and would avoid it if possible. In `hydra` when we create a client, we can choose its ID right? Can't we write a semantic ID based on the relation ID (and other stuff if necessary)?

Unfortunately, there is no way in hydra to choose the client_id. The reasoning behind it is that they create client_ids using some hash function to optimize SQL queries for deployments with millions (?) of clients. I tried to think of a way to avoid having to use the peer relation but I don't think we can (we could do a full diff, ie check the relations and compare them with the clients in the database, but that seems way worse).

gruyaume:
> > ```
> > * The question of the relation broken is a good one. I'm not a fan of having to maintain a mapping like this one and would avoid it if possible. In `hydra` when we create a client, we can choose its ID right? Can't we write a semantic ID based on the relation ID (and other stuff if necessary)?
> > ```
> 
> Unfortunately, there is no way in hydra to choose the client_id. The reasoning behind it is that they create client_ids using some hash function to optimize SQL queries for deployments with millions (?) of clients. I tried to think of a way to avoid having to use the peer relation but I don't think we can (we could do a full diff, ie check the relations and compare them with the clients in the database, but that seems way worse).

The API documentation seems to suggest that one can decide their ID (ref: https://www.ory.sh/docs/hydra/cli/hydra-clients-create). I haven't tried it so I don't know if it works.

gruyaume:
> > > ```
> > > * The question of the relation broken is a good one. I'm not a fan of having to maintain a mapping like this one and would avoid it if possible. In `hydra` when we create a client, we can choose its ID right? Can't we write a semantic ID based on the relation ID (and other stuff if necessary)?
> > > ```
> > 
> > 
> > Unfortunately, there is no way in hydra to choose the client_id. The reasoning behind it is that they create client_ids using some hash function to optimize SQL queries for deployments with millions (?) of clients. I tried to think of a way to avoid having to use the peer relation but I don't think we can (we could do a full diff, ie check the relations and compare them with the clients in the database, but that seems way worse).
> 
> [Ignore this message] The API documentation seems to suggest that one can decide their ID (ref: https://www.ory.sh/docs/hydra/cli/hydra-clients-create). I haven't tried it so I don't know if it works. Ahh it also seems like the expected format should be UUID V4 (ref: [ory/hydra#3154](https://github.com/ory/hydra/pull/3154)) :(


So yes it seems like we can't set the ID. Unless there's a metadata field in the `client create` where we could add the relation ID, I don't see a better approach than this mapping. Now the question is where to place this mapping. As you mentioned, peer relation data can be a good approach.

I'd recommend also considering secrets. You can use the relation ID as the secret label.

Creating the secret:
```python
content = {
            'relation_id': 32,
            'hydra_client_id': "abcd-efgh-1234",
        }
secret = self.app.add_secret(content, label='<relation ID>')
```

Fetching the secret:

```python
secret = self.model.get_secret(label='<relation ID>')
content = secret.get_content()
```

nsklikas:
> 
> So yes it seems like we can't set the ID. Unless there's a metadata field in the `client create` where we could add the relation ID, I don't see a better approach than this mapping. Now the question is where to place this mapping. As you mentioned, peer relation data can be a good approach. I'd recommend also considering secrets

There is a metadata field and I store the relation_id there, but you can't query on that (and even if you could it wouldn't be and indexed query). 

I would like to point out that since we are only going to store clients that were created from a relation, we know that the size of a mapping cannot get too big.

Using secrets is an interesting approach, but I think that we should avoid it. AFAIK currently juju secrets are stored in mongo but  in the future they will be moved to a secrets vault, I don't think we would want to store a simple mapping like this in a vault. Furthermore, semantically it does not make sense to use secrets like that (which could mean that this may break in other ways as well in the future).

I have already created a PR for using the peer relation (#33), so please let's move this discussion there.

tests/unit/test_oauth_provider.py
```diff
@@ -0,0 +1,200 @@
...
+        "requirer",
+        {
+            "redirect_uri": "https://oidc-client.com/callback",
+            "scope": "openid email",
+            "grant_types": '["authorization_code"]',
+            "audience": "[]",
+            "token_endpoint_auth_method": "client_secret_basic",
+        },
+    )
+    assert len(harness.charm.events) == 1
```
gruyaume, line 191: Your test should focus on validating 1 behavior. I'd remove these 2 asserts

lib/charms/hydra/v0/oauth.py
```diff
@@ -0,0 +1,672 @@
...
+            return
+
+        data = _dump_data(client_config.to_dict(), OAUTH_REQUIRER_JSON_SCHEMA)
+        relation.data[self.model.app].update(data)
+
+    def get_provider_info(self) -> Optional[Dict]:
+        """Get the provider information from the databag."""
+        if len(self.model.relations) == 0:
+            return None
+        relation = self.model.get_relation(self._relation_name)
```
gruyaume, line 369: Don't we want to validate for `TooManyRelatedAppsError`? In any case I'd put all that validation in a separate method, we already have the same try/except in 2 places in this file

src/charm.py
```diff
@@ -303,20 +328,180 @@ def _on_admin_ingress_ready(self, event: IngressPerAppReadyEvent) -> None:
...
+            "client",
+            "--endpoint",
+            self.hydra_admin_url,
+            "--format",
+            "json",
+        ]
+
+    def create_client(
+        self, client_config: ClientConfig, metadata: Optional[Union[Dict, str]] = None
+    ) -> Optional[Dict]:
```
gruyaume, line 465: Why is this an optional return?

src/charm.py
```diff
@@ -303,20 +328,180 @@ def _on_admin_ingress_ready(self, event: IngressPerAppReadyEvent) -> None:
...
+        """Create an oauth2 client."""
+        cmd = self._client_cmd_prefix("create") + self._client_config_to_cmd(
+            client_config, metadata
+        )
+
+        stdout, _ = self._run_cmd(cmd)
+        stdout = json.loads(stdout)
```
gruyaume, line 472: I'd use a different variable, the type/content is different from the previous `stdout`

lib/charms/hydra/v0/oauth.py
```diff
@@ -0,0 +1,672 @@
...
+            # Create a new client
+            self.on.client_created.emit(
+                redirect_uri, scope, grant_types, audience, token_endpoint_auth_method, relation_id
+            )
+
+    def _get_secret_label(self, relation: Relation) -> str:
+        return f"client_secret_{relation.id}"
+
+    def _on_relation_departed(self, event: RelationDepartedEvent) -> None:
+        # The relation has been departed, delete the juju secret
```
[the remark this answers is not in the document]
nsklikas: It is comment about the next couple of lines, not the whole function. I would rather keep it this way, if it's ok with you.

tests/unit/test_charm.py
```diff
@@ -41,6 +45,13 @@ def setup_ingress_relation(harness, type):
     return relation_id
 
 
+def setup_oauth_relation(harness):
```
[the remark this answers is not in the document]
nsklikas: We don't have type annotations in our tests, we should add to all of them (if that's what we want)

tests/unit/test_oauth_provider.py
nsklikas: If those asserts were to fail that means that the rest of the test will not work, as the client will not be created thus making it impossible to remove the client_secret. They are there so that the dev will get a clearer message in case something goes wrong.

gruyaume: Sure but the goal of the test is to validate 1 behavior and this assert validates "something else". If you really want to validate this behavior you may have a test specifically for it but I'd discourage testing for the same thing in multiple tests.

tests/unit/test_oauth_provider.py
```diff
@@ -0,0 +1,200 @@
...
+    assert isinstance(harness.charm.events[0], ClientCreatedEvent)
+
+    harness.remove_relation(relation_id)
+
+    assert len(harness.charm.events) == 2
```
gruyaume, line 173: Do we really want to num of events to be 2? We don't really care right? What we care about in this test is that the `ClientDeletedEvent` was triggered.

lib/charms/hydra/v0/oauth.py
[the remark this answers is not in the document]
gruyaume: If you think it is valuable, sure keep it. But the use of comments/docstrings is pretty inconsistent. For example what drives us to want a comment here but none in `get_client_credentials`?

tests/unit/test_charm.py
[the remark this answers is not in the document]
nsklikas: Very well, I will add the types here and create a card for adding type annotations to the rest of the tests

tests/unit/test_oauth_provider.py
gruyaume: Well it's a bit more nuanced than this. Take this example

```python
def test_xyz():
    """
    Given: Bank Account with initial balance of 10$
    When: Transfer 5$
    Then: Balance is 15$
    """
    bank_account = BankAccount(balance=10)
    
    bank_account.transfer(5)

    assert bank_account.balance == 15
```

It's not so much that this test is bad. Indeed if the transfer is of 5$, we expect the balance to be 15$. The problem is that it's not asserting what we really want to assert which is that the account balance is the sum of the initial balance and the transfer amount. So a better way to write it is:

```python
def test_xyz():
    """
    Given: Bank Account with initial balance
    When: Transfer amount
    Then: Balance is sum of initial balance and transfer amount
    """
    initial_balance = 10
    transfer_amount = 5
    bank_account = BankAccount(balance=initial_balance)
    
    bank_account.transfer(transfer_amount)

    assert bank_account.balance == initial_balance + transfer_amount
```

nsklikas:
I tried to refactor the tests and improve their naming. Please have another go at reviewing it.

tests/unit/test_oauth_requirer.py
```diff
@@ -0,0 +1,212 @@
...
+    def _record_event(self, event: EventBase) -> None:
+        self.events.append(event)
+
+
+def test_data_in_relation_bag(harness: Harness) -> None:
+    relation_id = harness.add_relation("oauth", "provider")
+
+    relation_data = harness.get_relation_data(relation_id, harness.model.app.name)
+
+    assert _load_data(relation_data) == CLIENT_CONFIG
```
gruyaume, line 78: You don't want this sort of coupling with your source code. `_load_data` is a private method and if for whatever reason it changes, all tests would crash.

gruyaume (commented):
Good job on the unit tests, they look much cleaner. I just gave 1 comment on coupling

natalian98 (approved):

gruyaume (approved):
Good job!

Issue #31, opened 2023-03-09
Hydra fails to deploy in a bundle when traefik-k8s is not ready


natalian98:
### Bug Description

When deploying a bundle with hydra and `endpoint-info` relation with kratos, hydra may end up in error state.

This happens because on `relation-created` hydra is trying to send traefik-k8s's urls when the `hydra:admin-ingress` and `hydra:public-ingress` relations are place, but those urls are still `None` as traefik is getting deployed.
The solution is to check if traefik is ready rather than just checking if the relation exists - the fact it was created doesn't imply its relation data is ready.

### To Reproduce

`juju deploy ./bundle.yaml`

### Environment

bundle.yaml
```
bundle: kubernetes
name: iam
applications:
  hydra:
    charm: hydra
    channel: edge
    scale: 1
    series: jammy
    trust: true
  kratos:
    charm: kratos
    channel: edge
    scale: 1
    series: jammy
    trust: true
  kratos-external-idp-integrator:
    charm: kratos-external-idp-integrator
    channel: edge
    scale: 1
    series: jammy
  postgresql-k8s:
    charm: postgresql-k8s
    channel: edge
    series: jammy
    scale: 1
    trust: true
  traefik-admin:
    charm: traefik-k8s
    channel: edge
    series: focal
    scale: 1
  traefik-public:
    charm: traefik-k8s
    channel: edge
    series: focal
    scale: 1
relations:
  - [hydra:pg-database, postgresql-k8s:database]
  - [kratos:pg-database, postgresql-k8s:database]
  - [kratos:endpoint-info, hydra:endpoint-info]
  - [kratos-external-idp-integrator:kratos-external-idp, kratos:kratos-external-idp]
  - [hydra:admin-ingress, traefik-admin:ingress]
  - [hydra:public-ingress, traefik-public:ingress]
  - [kratos:admin-ingress, traefik-admin:ingress]
  - [kratos:public-ingress, traefik-public:ingress]
```

### Relevant log output

```shell
unit-hydra-0: 13:23:51 INFO juju.worker.uniter awaiting error resolution for "relation-created" hook
unit-hydra-0: 13:23:52 ERROR unit.hydra/0.juju-log endpoint-info:5: Uncaught exception while in charm code:
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-hydra-0/charm/./src/charm.py", line 322, in <module>
    main(HydraCharm)
[...]

PR #32, opened 2023-03-09
fix: check if traefik is ready instead of checking if relation is created

natalian98:
fixes https://github.com/canonical/hydra-operator/issues/31


nsklikas:
Also can you write a test that would fail without your fix?

nsklikas (approved):

PR #33, opened 2023-03-10
Delete clients when oauth relation is broken

nsklikas:
Add peer relation that will be used for deleting clients when the `oauth` relation is broken

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -41,7 +41,14 @@ from ops.charm import (
     WorkloadEvent,
 )
 from ops.main import main
-from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, ModelError, WaitingStatus, Relation
+from ops.model import (
+    ActiveStatus,
+    BlockedStatus,
+    MaintenanceStatus,
+    ModelError,
+    Relation,
+    WaitingStatus,
+)
 from ops.pebble import ChangeError, ExecError, Layer
 
 from hydra_cli import HydraCLI
--- a/src/charm.py
+++ b/src/charm.py
@@ -6,9 +6,10 @@
 
 """A Juju charm for Ory Hydra."""
 
+import json
 import logging
 from os.path import join
-from typing import Any
+from typing import Any, Dict
 
 from charms.data_platform_libs.v0.data_interfaces import (
     DatabaseCreatedEvent,
@@ -456,6 +457,7 @@ class HydraCharm(CharmBase):
         client = self._get_oauth_relation_peer_data(event.relation_id)
         if not client:
             logger.error("No client found in peer data")
+            return
 
         try:
             self._hydra_cli.delete_client(client["client_id"])
--- a/src/charm.py
+++ b/src/charm.py
@@ -42,7 +42,7 @@ from ops.charm import (
 )
 from ops.main import main
 from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, ModelError, WaitingStatus
-from ops.pebble import ChangeError, Error, ExecError, Layer
+from ops.pebble import ChangeError, ExecError, Layer
 
 from hydra_cli import HydraCLI
 
@@ -409,9 +409,6 @@ class HydraCharm(CharmBase):
             )
         except ExecError as err:
             logger.error(f"Exited with code: {err.exit_code}. Stderr: {err.stderr}")
-            return
-        except Error as e:
-            logger.error(f"Something went wrong when trying to run the command: {e}")
             logger.info("Deferring the event")
             event.defer()
             return
@@ -437,21 +434,15 @@ class HydraCharm(CharmBase):
             )
         except ExecError as err:
             logger.error(f"Exited with code: {err.exit_code}. Stderr: {err.stderr}")
-            return
-        except Error as e:
-            logger.error(f"Something went wrong when trying to run the command: {e}")
             logger.info("Deferring the event")
             event.defer()
             return
 
     def _on_client_deleted(self, event: ClientDeletedEvent) -> None:
-        if not self._container.can_connect():
-            event.defer()
+        if not self.unit.is_leader():
             return
 
-        try:
-            self._container.get_service(self._container_name)
-        except (ModelError, RuntimeError):
+        if not self._hydra_service_is_running:
             event.defer()
             return
 
@@ -465,9 +456,6 @@ class HydraCharm(CharmBase):
             self._hydra_cli.delete_client(client["client_id"])
         except ExecError as err:
             logger.error(f"Exited with code: {err.exit_code}. Stderr: {err.stderr}")
-            return
-        except Error as e:
-            logger.error(f"Something went wrong when trying to run the command: {e}")
             logger.info("Deferring the event")
             event.defer()
             return
--- a/src/charm.py
+++ b/src/charm.py
@@ -212,11 +212,11 @@ class HydraCharm(CharmBase):
         return self.model.get_relation(PEER)
 
     def _set_peer_data(self, key: str, data: Dict) -> None:
-        """Put information into the peer data bucket instead of `StoredState`."""
+        """Put information into the peer data bucket."""
         self._peers.data[self.app][key] = json.dumps(data)
 
     def _get_peer_data(self, key: str) -> Dict:
-        """Retrieve information from the peer data bucket instead of `StoredState`."""
+        """Retrieve information from the peer data bucket."""
         if not self._peers:
             return {}
         data = self._peers.data[self.app].get(key, "")
--- a/src/charm.py
+++ b/src/charm.py
@@ -419,7 +419,6 @@ class HydraCharm(CharmBase):
             event.defer()
             return
 
-        # TODO: If this fails emit an event for reconciliation
         self._set_oauth_relation_peer_data(event.relation_id, dict(client_id=client["client_id"]))
         self.oauth.set_client_credentials_in_relation_data(
             event.relation_id, client["client_id"], client["client_secret"]
@@ -459,7 +458,6 @@ class HydraCharm(CharmBase):
             event.defer()
             return
 
-        # TODO: If this fails emit an event for reconciliation
         client = self._get_oauth_relation_peer_data(event.relation_id)
         if not client:
             logger.error("No client found in peer data")
@@ -473,7 +471,6 @@ class HydraCharm(CharmBase):
             event.defer()
             return
 
-        # TODO: If this fails emit an event for reconciliation
         self._pop_oauth_relation_peer_data(event.relation_id)
 
     def _update_endpoint_info(self) -> None:
--- a/src/charm.py
+++ b/src/charm.py
@@ -16,7 +16,12 @@ from charms.data_platform_libs.v0.data_interfaces import (
     DatabaseRequires,
 )
 from charms.hydra.v0.hydra_endpoints import HydraEndpointsProvider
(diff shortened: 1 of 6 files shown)


src/charm.py
```diff
@@ -170,6 +199,44 @@ def _run_sql_migration(self, set_timeout: bool) -> None:
...
+    def _oauth_relation_peer_data_key(self, relation_id: int) -> str:
+        return f"oauth_{relation_id}"
+
+    @property
+    def _peers(self):
+        """Fetch the peer relation."""
+        return self.model.get_relation(PEER)
+
+    def _set_peer_data(self, key: str, data: Dict) -> None:
+        """Put information into the peer data bucket instead of `StoredState`."""
```
gruyaume, line 211:
```suggestion
        """Put information into the peer data bucket."""
```

src/charm.py
```diff
@@ -170,6 +199,44 @@ def _run_sql_migration(self, set_timeout: bool) -> None:
...
+
+    def _set_oauth_relation_peer_data(self, relation_id: int, data: Dict) -> None:
+        key = self._oauth_relation_peer_data_key(relation_id)
+        self._set_peer_data(key, data)
+
+    def _get_oauth_relation_peer_data(self, relation_id: int) -> None:
+        key = self._oauth_relation_peer_data_key(relation_id)
+        return self._get_peer_data(key)
+
+    def _pop_oauth_relation_peer_data(self, relation_id: int) -> None:
```
gruyaume, line 236: The return typing doesn't match the actual return

src/charm.py
```diff
@@ -170,6 +199,44 @@ def _run_sql_migration(self, set_timeout: bool) -> None:
...
+    def _peers(self):
+        """Fetch the peer relation."""
+        return self.model.get_relation(PEER)
+
+    def _set_peer_data(self, key: str, data: Dict) -> None:
+        """Put information into the peer data bucket instead of `StoredState`."""
+        self._peers.data[self.app][key] = json.dumps(data)
+
+    def _get_peer_data(self, key: str) -> Dict:
+        """Retrieve information from the peer data bucket instead of `StoredState`."""
```
gruyaume, line 215:
```suggestion
        """Retrieve information from the peer data bucket."""
```

src/charm.py
```diff
@@ -302,20 +369,212 @@ def _on_admin_ingress_ready(self, event: IngressPerAppReadyEvent) -> None:
...
+            authorization_endpoint=join(self.public_ingress.url, "oauth2/auth"),
+            token_endpoint=join(self.public_ingress.url, "oauth2/token"),
+            introspection_endpoint=join(self.admin_ingress.url, "admin/oauth2/introspect"),
+            userinfo_endpoint=join(self.public_ingress.url, "userinfo"),
+            jwks_endpoint=join(self.public_ingress.url, ".well-known/jwks.json"),
+            scope=SUPPORTED_SCOPES,
+        )
+
+
+class HydraCLI:
```
gruyaume, line 491: Could/should we place this class in a different file?

src/charm.py
```diff
@@ -170,6 +199,44 @@ def _run_sql_migration(self, set_timeout: bool) -> None:
...
+        if not self._peers:
+            return {}
+        data = self._peers.data[self.app].pop(key, "")
+        return json.loads(data) if data else {}
+
+    def _set_oauth_relation_peer_data(self, relation_id: int, data: Dict) -> None:
+        key = self._oauth_relation_peer_data_key(relation_id)
+        self._set_peer_data(key, data)
+
+    def _get_oauth_relation_peer_data(self, relation_id: int) -> None:
```
gruyaume, line 232: The return typing doesn't match the actual return

src/charm.py
```diff
@@ -302,20 +369,212 @@ def _on_admin_ingress_ready(self, event: IngressPerAppReadyEvent) -> None:
...
+        self.oauth.set_client_credentials_in_relation_data(
+            event.relation_id, client["client_id"], client["client_secret"]
+        )
+
+    def _on_client_changed(self, event: ClientChangedEvent) -> None:
+        if not self._container.can_connect():
+            event.defer()
+            return
+
+        try:
```
gruyaume, line 428: I see this try/except in a couple of places. You may want to create a `_service_is_running(self, service_name: str)` method and call it here:

```python
if not self._service_is_running(service_name):
    event.defer()
    return
```

src/charm.py
```diff
@@ -302,20 +369,212 @@ def _on_admin_ingress_ready(self, event: IngressPerAppReadyEvent) -> None:
...
+            event.defer()
+            return
+
+        client = self._get_oauth_relation_peer_data(event.relation_id)
+        if not client:
+            logger.error("No client found in peer data")
+
+        try:
+            self._hydra_cli.delete_client(client["client_id"])
+        except ExecError as err:
```
gruyaume, line 465: There are lots of these error handlings, could that be done in the hydra class? (of course excluding the `defer`)

src/charm.py
```diff
@@ -302,20 +369,212 @@ def _on_admin_ingress_ready(self, event: IngressPerAppReadyEvent) -> None:
...
+        logger.info(f"Successfully updated client: {client_config.client_id}")
+        return json.loads(stdout)
+
+    def delete_client(self, client_id: str) -> Optional[Dict]:
+        """Delete one or more oauth2 client."""
+        cmd = self._client_cmd_prefix("delete")
+        cmd.append(client_id)
+
+        stdout, _ = self._run_cmd(cmd)
+        logger.info(f"Successfully deleted clients: {stdout}")
```
gruyaume, line 568:
```suggestion
        logger.info(f"Successfully deleted client: {stdout}")
```

src/charm.py
```diff
@@ -302,20 +369,212 @@ def _on_admin_ingress_ready(self, event: IngressPerAppReadyEvent) -> None:
...
+            client_config, metadata
+        )
+        cmd.append(client_config.client_id)
+
+        stdout, _ = self._run_cmd(cmd)
+        logger.info(f"Successfully updated client: {client_config.client_id}")
+        return json.loads(stdout)
+
+    def delete_client(self, client_id: str) -> Optional[Dict]:
+        """Delete one or more oauth2 client."""
```
gruyaume, line 563:
```suggestion
        """Delete one oauth2 client."""
```

src/charm.py
```diff
@@ -302,20 +369,212 @@ def _on_admin_ingress_ready(self, event: IngressPerAppReadyEvent) -> None:
...
+        """Convert a ClientConfig object to a list of parameters."""
+        flags = [
+            "--grant-type",
+            ",".join(client_config.grant_types or ["authorization_code", "refresh_token"]),
+            "--response-type",
+            "code",
+        ]
+
+        if client_config.scope:
+            for s in client_config.scope.split(" "):
```
gruyaume, line 510: use `scope` instead of `s`

src/charm.py
```diff
@@ -302,20 +369,212 @@ def _on_admin_ingress_ready(self, event: IngressPerAppReadyEvent) -> None:
...
+
+class HydraCLI:
+    """Helper object for running hydra CLI commands."""
+
+    def __init__(self, hydra_admin_url: str, container: Container):
+        self.hydra_admin_url = hydra_admin_url
+        self.container = container
+
+    def _client_config_to_cmd(
+        self, client_config: ClientConfig, metadata: Optional[Union[Dict, str]] = None
```
gruyaume, line 499: Is there a reason why we want to accept both string and dit for the metadata? Seems simpler to settle on one

src/charm.py
```diff
@@ -302,20 +369,212 @@ def _on_admin_ingress_ready(self, event: IngressPerAppReadyEvent) -> None:
...
             logger.info("This app no longer has ingress")
 
         self._handle_status_update_config(event)
         self._update_hydra_endpoints_relation_data(event)
+        self._update_endpoint_info()
+
+    def _on_oauth_relation_created(self, event: RelationCreatedEvent) -> None:
+        self._update_endpoint_info()
+
+    def _on_client_created(self, event: ClientCreatedEvent) -> None:
```
gruyaume, line 393: There seems to be missing validations that the unit is the leader. For instance we probably don't want all units to create clients and store the mapping in the peer relation data.

gruyaume: There also seems to be lacking validation that the peer relation was created before adding/removing data from it.

src/charm.py
nsklikas: I suppose you mean the HydraCLI class. 
I was thinking about this, the thing is we would still have to either throw an error or return an error value in case something goes wrong. I don't think it would simplify the code.

src/charm.py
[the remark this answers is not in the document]
nsklikas: I agree that if this fails, it could cause unexpected errors to happen.

BUT going through all the relations and verifying that the corresponding client exists can be a very expensive operation, depending on how the admin chooses to use the platform. We would have to get the entire client table for the database and compare that with the contents of the juju relation.

I think that the best we can do is:
- Provide a sanitization command that would do that checking.
- Catch any errors that may be thrown when popping from the peer relation and log a descriptive error, urging the admin to run the sanitization command.

src/charm.py
nsklikas: I looked at other charms code to do this and they weren't checking if the peer relation was ready.
How would you do that? Is something like this enough?
```python
if not self.model.get_relation(PEER):
    return
```

src/charm.py
nsklikas: Yes, the class got to big. I think it makes sense to move it to a different file.

src/charm.py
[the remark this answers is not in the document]
nsklikas: After some more thought on this, I realized that, while this is a very valid concern, in general we don't handle failures when trying to get/set relation data (correct me if I'm wrong). I don't see why this case is any different.

To fix this we could wrap this in a `try/except` block and fire an event for reconciling the peer data, but we need to first understand when that can actually fail and what it means that it failed (will juju try to fire the event again?) and if we actually need the user to do something or not (can be just a temporary error?).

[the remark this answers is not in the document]
massigori: Taking a step back on this one I consider deleting clients when a relation is broken as something that is _nice to have_ form a security/operational point of view, but not _essential_ (unless the prior credentials are leaked)

For the time being, and until we are able to observe the real behaviour of the charm in the field, I suggest that in the event of an error we throw an informative message that instructs the admin to check the active clients in hydra, either through the charm actions or the cli

I would also open an issue on this repo detailing the possible error conditions and further work required

src/charm.py
gruyaume: I thing `get_relation` throws an error when the relation name doesn't exist but yes somehting along those lines

tests/unit/test_charm.py
```diff
@@ -55,6 +55,12 @@ def setup_oauth_relation(harness: Harness) -> Tuple[int, str]:
     return relation_id, app_name
 
 
+def setup_peer_relation(harness):
```
natalian98, line 58: Add typing

src/charm.py
```diff
@@ -195,6 +211,46 @@ def _run_sql_migration(self, set_timeout: bool) -> None:
...
+
+    def _get_peer_data(self, key: str) -> Dict:
+        """Retrieve information from the peer data bucket."""
+        if not (peers := self._peers):
+            return {}
+        data = peers.data[self.app].get(key, "")
+        return json.loads(data) if data else {}
+
+    def _pop_peer_data(self, key: str) -> Dict:
+        """Retrieve information from the peer data bucket."""
```
gruyaume, line 236: The docstring doesn't match the method name

gruyaume (approved):

PR #37, opened 2023-03-16
Add actions

nsklikas:
Adds juju actions for managing clients, revoking a client's tokens and rotating the id token signing key.

I had to access the hydra CLI, so I based this PR on top of https://github.com/canonical/hydra-operator/pull/29

TODO:
- [x] Add unit tests
- [x] Add integration tests
- [x] Should we add action for running `hydra janitor`?
- [ ] Test the `list-oauth-clients` command in deployments with many clients


nsklikas:
After thinking about it, I think it is best not to have an action calling the janitor command:
1) If the user wants to run the `hydra janitor` command manually, they would want to see the output of the command (live), AFAIK juju provides no way of doing this.
2) If the user wants to run the command periodically, they could use a cronjob or maybe we could provide a config option for doing this, but I don't see a reason to have a cronjob calling a juju action periodically.

unit-requirements.txt
```diff
@@ -2,4 +2,5 @@ pytest
 pytest-mock
 coverage[toml]
 requests
+ipdb
```
gruyaume, line 5: Do we need this?

src/charm.py
```diff
@@ -396,6 +426,202 @@ def _on_client_changed(self, event: ClientChangedEvent) -> None:
             event.defer()
             return
 
+    def _on_create_oauth_client_action(self, event: ActionEvent) -> None:
```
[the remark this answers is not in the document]
nsklikas: I don't see any reason for stopping the user from running these actions from any unit

src/hydra_cli.py
```diff
@@ -21,27 +21,48 @@ def __init__(self, hydra_admin_url: str, container: Container):
         self.hydra_admin_url = hydra_admin_url
         self.container = container
 
-    def _client_config_to_cmd(
-        self, client_config: ClientConfig, metadata: Optional[Dict] = None
+    def _dump_list_or_dict(self, o: Union[Dict, List, str]) -> str:
```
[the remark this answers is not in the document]
nsklikas: To avoid having to repeat logic, as well as making it easier to add more params if needed.

unit-requirements.txt
nsklikas: We set `PYTHONBREAKPOINT=ipdb.set_trace`, shouldn't we install `ipdb` to enable setting breakpoints?

gruyaume: I hadn't seen this line, so maybe you do!

src/hydra_cli.py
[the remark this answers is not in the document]
gruyaume: If we have 1 method per type we wouldn't duplicate the logic and we'd avoid validating the `is_instance`.

src/charm.py
```diff
@@ -396,6 +427,219 @@ def _on_client_changed(self, event: ClientChangedEvent) -> None:
...
+        except Error as e:
+            event.fail(f"Something went wrong when trying to run the command: {e}")
+            return
+
+        event.log("Successfully created new key")
+        event.set_results({"new-key-id": jwk["keys"][0]["kid"]})
+
+    def _is_oauth_relation_client(self, client: Dict) -> bool:
+        """Check whether a client is managed from an oauth relation."""
+        return client.get("metadata", {}).get("relation_id")
```
gruyaume, line 641: The `relation_id` is a bool?

actions.yaml
```diff
@@ -1,2 +1,88 @@
...
+      type: string
+  required: ["client-id"]
+rotate-key:
+  description: Rotate the jwk used for signing tokens
+  params:
+    alg:
+      description: The algorithm that should be used
+      type: string
+      default: RS256
+      enumerate: [RS256, RS512, ES256, ES512, EdDSA]
```
natalian98, line 88: Does Ory Ory support RS512? It's not mentioned in their [docs](https://www.ory.sh/docs/hydra/cli/hydra-keys-create)

src/charm.py
```diff
@@ -396,6 +427,219 @@ def _on_client_changed(self, event: ClientChangedEvent) -> None:
             event.defer()
             return
 
+    def _on_create_oauth_client_action(self, event: ActionEvent) -> None:
+        if not self._hydra_service_is_running:
+            event.fail("Service is not ready.")
```
natalian98, line 432: Can we consider a more indicative message, for example `Service is not ready. Please re-run the action when the charm is active`?

actions.yaml
nsklikas: That's interesting, we use the [hydra create jwk](https://www.ory.sh/docs/hydra/cli/hydra-create-jwk) command which allows it.

natalian98 (approved):

gruyaume (commented):
Minor comment. Good job on the unit tests btw they look pretty clean

actions.yaml
```diff
@@ -1,2 +1,88 @@
...
+      type: array
+      default: ["authorization_code"]
+    redirect-uris:
+      description: A list with the client's redirect_uri
+      type: array
+    response-types:
+      description: A list with the allowed response types for the client
+      type: array
+      default: ["code"]
+    scope:
```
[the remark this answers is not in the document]
nsklikas: I agree that it is not intuitive to use singular, but in the auth request the client will request for `scope` (https://www.rfc-editor.org/rfc/rfc6749#section-4.1.1). 

We should use whatever will make most sense to the user, @massigori any thoughts on this?

gruyaume (dismissed):
Approved assuming the question about scope singular/plural is resolved. Good job :)

natalian98 (approved):

PR #45, opened 2023-03-31
fix potential bug of referencing relation name

bencekov:
I realise that this could be the bad platform to request a change for a charm library. Let me know and I'll close this PR.

Please check if this is potentially a bug in the hydra_endpoints library. if we insist on using the relation name in the constant, then it's likely a bad idea to make it an argument for the constructor.

### DIFF
--- a/lib/charms/hydra/v0/hydra_endpoints.py
+++ b/lib/charms/hydra/v0/hydra_endpoints.py
@@ -94,7 +94,7 @@ class HydraEndpointsProvider(Object):
         if not self._charm.unit.is_leader():
             return
 
-        relations = self.model.relations[RELATION_NAME]
+        relations = self.model.relations[self._relation_name]
         for relation in relations:
             relation.data[self._charm.app].update(
                 {


nsklikas (changes requested):
Good catch! Please update the library version as well (https://github.com/canonical/hydra-operator/pull/45/files#diff-4c3ead233e9ad9981244f226aecede4323ae42a82b2db34893e50f0cf59d45ecR56)

lib/charms/hydra/v0/hydra_endpoints.py
```diff
@@ -94,7 +94,7 @@ def send_endpoint_relation_data(self, admin_endpoint: str, public_endpoint: str)
         if not self._charm.unit.is_leader():
             return
 
-        relations = self.model.relations[RELATION_NAME]
+        relations = self.model.relations[self._relation_name]
```
nsklikas, line 97:
```suggestion
        relations = self.model.relations.get(self._relation_name)
        if not relations:
            return
```

Let's also check that the relation exists.

PR #65, opened 2023-05-15
Use service to communicate with other components

nsklikas:
There is no reason to use the ingress since all of our components will be in the same k8s cluster, in the same model.


bencekov (approved):
Looks good!

PR #66, opened 2023-05-17
Use ROCK image

nsklikas:
Switch metadata image to canonical ROCK

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -16,7 +16,7 @@ resources:
   oci-image:
     type: oci-image
     description: OCI image for hydra container
-    upstream-source: oryd/hydra:v2.0.3
+    upstream-source: ghcr.io/canonical/hydra:2.1.1
 requires:
   pg-database:
     interface: postgresql_client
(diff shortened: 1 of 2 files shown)


metadata.yaml
gruyaume, line 1: The integration tests are currently failing, `cat` doesn't exist anymore :cat:

nsklikas: That's unfortunate. We can't use `kubectl cp`, because it needs `tar`. The only way that I can think of for exporting the file would be to curl the pebble files HTTP API, but I am not certain what that request would have to look like.

I am going to remove that step for now.

gruyaume: Why do we care about this file? Can't we validate hydra CLI calls instead?

gruyaume (dismissed):

metadata.yaml
nsklikas: This is not a test, it's a debugging step. The problem is that it's missing the `if: failure` part and it always runs.

natalian98 (approved):

PR #67, opened 2023-05-18
IAM-259-Add COS Integration to Hydra

bencekov:
Integrating Hydra charm with COS components: Prometheus and Loki

cos integrations manual testing instructions:

These instructions presuppose that you have a juju cluster with working hydra, prometheus-k8s, and loki-k8s integrated with hydra. Hydra charm uses the default relation names.

Prometheus:
curl 'prometheus_url:9090/api/v1/query?query=http_requests_total&app=hydra&code=200&endpoint=/admin/health/alive' | yq
If you get a list of metrics, then it worked.

Loki
curl loki_url:3100/loki/api/v1/label/juju_unit/values
If your unit shows up in the results, it should be good.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -131,7 +131,7 @@ class HydraCharm(CharmBase):
             relation_name=self._prometheus_scrape_relation_name,
             jobs=[
                 {
-                    "metrics_path": "admin/metrics/prometheus",
+                    "metrics_path": "/admin/metrics/prometheus",
                     "static_configs": [
                         {
                             "targets": [f"*:{HYDRA_ADMIN_PORT}"],
--- a/src/charm.py
+++ b/src/charm.py
@@ -252,9 +252,12 @@ class HydraCharm(CharmBase):
     def _log_level(self) -> str:
         return self.config["log_level"]
 
-    @property
-    def _log_level_valid(self) -> bool:
-        return self._log_level in LOG_LEVELS
+    def _validate_config_log_level(self) -> bool:
+        is_valid = self._log_level in LOG_LEVELS
+        if not is_valid:
+            logger.info(f"Invalid configuration value for log_level: {self._log_level}")
+            self.unit.status = BlockedStatus("Invalid configuration value for log_level")
+        return is_valid
 
     def _render_conf_file(self) -> str:
         """Render the Hydra configuration file."""
@@ -342,6 +345,9 @@ class HydraCharm(CharmBase):
             self.unit.status = WaitingStatus("Waiting to connect to Hydra container")
             return
 
+        if not self._validate_config_log_level():
+            return
+
         self.unit.status = MaintenanceStatus("Configuring resources")
 
         current_layer = self._container.get_plan()
@@ -360,11 +366,6 @@ class HydraCharm(CharmBase):
             event.defer()
             self.unit.status = WaitingStatus("Waiting for Hydra service")
             logger.info("Hydra service is absent. Deferring the event.")
-            return
-
-        if not self._log_level_valid:
-            logger.info(f"Invalid configuration value for log_level: {self._log_level}")
-            self.unit.status = BlockedStatus("Bad configuration value for log_level")
             return
 
         if not self.model.relations[self._db_relation_name]:
@@ -410,6 +411,10 @@ class HydraCharm(CharmBase):
             self.unit.status = WaitingStatus("Waiting to connect to Hydra container")
             return
 
+        if not self._validate_config_log_level():
+            event.defer()
+            return
+
         self.unit.status = MaintenanceStatus(
             "Configuring container and resources for database connection"
         )
@@ -418,11 +423,6 @@ class HydraCharm(CharmBase):
             event.defer()
             self.unit.status = WaitingStatus("Waiting for Hydra service")
             logger.info("Hydra service is absent. Deferring the event.")
-            return
-
-        if not self._log_level_valid:
-            logger.info(f"Invalid configuration value for log_level: {self._log_level}")
-            self.unit.status = BlockedStatus("Bad configuration value for log_level")
             return
 
         logger.info("Updating Hydra config and restarting service")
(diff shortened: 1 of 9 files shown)


gruyaume (dismissed):

src/charm.py
```diff
@@ -117,6 +125,28 @@ def __init__(self, *args: Any) -> None:
 
         self.endpoints_provider = HydraEndpointsProvider(self)
 
+        self.metrics_endpoint = MetricsEndpointProvider(
+            self,
+            relation_name=self._prometheus_scrape_relation_name,
```
natalian98, line 130: it's the default prometheus relation name so you don't need to specify it
```suggestion

```

src/charm.py
```diff
@@ -117,6 +125,28 @@ def __init__(self, *args: Any) -> None:
...
+                        }
+                    ],
+                }
+            ],
+        )
+
+        self.loki_consumer = LogProxyConsumer(
+            self,
+            log_files=[self._log_path],
+            relation_name=self._loki_push_api_relation_name,
```
natalian98, line 146: similar to above - `logging` is the default loki relation name so there is no need to provide it
```suggestion

```

src/charm.py
```diff
@@ -84,6 +87,11 @@ def __init__(self, *args: Any) -> None:
         self._db_name = f"{self.model.name}_{self.app.name}"
         self._db_relation_name = "pg-database"
         self._login_ui_relation_name = "ui-endpoint-info"
+        self._prometheus_scrape_relation_name = "metrics-endpoint"
+        self._loki_push_api_relation_name = "logging"
+        self._hydra_service_command = "hydra serve all"
```
natalian98, line 92: why do you put the serve command in a variable?

src/charm.py
```diff
@@ -316,6 +362,11 @@ def _handle_status_update_config(self, event: HookEvent) -> None:
             logger.info("Hydra service is absent. Deferring the event.")
             return
 
+        if not self._log_level_valid:
+            logger.info(f"Invalid configuration value for log_level: {self._log_level}")
+            self.unit.status = BlockedStatus("Bad configuration value for log_level")
+            return
+
```
natalian98, line 369: we could check if the log level value is valid before rendering the template - it shouldn't be rendered with a wrong config value

src/charm.py
```diff
@@ -369,6 +420,11 @@ def _on_database_created(self, event: DatabaseCreatedEvent) -> None:
             logger.info("Hydra service is absent. Deferring the event.")
             return
 
+        if not self._log_level_valid:
+            logger.info(f"Invalid configuration value for log_level: {self._log_level}")
+            self.unit.status = BlockedStatus("Bad configuration value for log_level")
```
natalian98, line 425: I would reformat the status message a little
```suggestion
            self.unit.status = BlockedStatus("Invalid configuration value for log_level")
```

src/charm.py
```diff
@@ -764,6 +820,10 @@ def _get_login_ui_endpoint_info(self, key: str) -> Optional[str]:
             logger.info("Too many ui-endpoint-info relations found")
         return None
 
+    def _promtail_error(self, event: PromtailDigestError):
```
natalian98, line 823:
```suggestion
    def _promtail_error(self, event: PromtailDigestError) -> None:
```

natalian98:
Can we have some example prometheus rule added? Sth like [here](https://github.com/jnsgruk/zinc-k8s-operator/blob/main/src/prometheus_alert_rules/zinc_missing.rule)

src/charm.py
bencekov: I'm gonna copy paste my answer here, because this issue is raised in multiple places:

The default relation name for the class we use to interact with loki isn't logging but log-proxy.
You can check at https://github.com/canonical/loki-k8s-operator/blob/main/lib/charms/loki_k8s/v0/loki_push_api.py line 491.
The documentation and the communication around the default relation name has been a bit of a mess with contradictory information.
As a consequence of this, I decided to always include the relation name. I also think this is a good idea because the default relation names can be updated in the libs anytime without us knowing.

src/charm.py
bencekov: I'm gonna copy paste my answer here, because this issue is raised in multiple places:

The default relation name for the class we use to interact with loki isn't logging but log-proxy.
You can check at https://github.com/canonical/loki-k8s-operator/blob/main/lib/charms/loki_k8s/v0/loki_push_api.py line 491.
The documentation and the communication around the default relation name has been a bit of a mess with contradictory information.
As a consequence of this, I decided to always include the relation name. I also think this is a good idea because the default relation names can be updated in the libs anytime without us knowing.

src/charm.py
bencekov: I try to keep the code in our charms uniform, and store the specifics in variables.

src/charm.py
```diff
@@ -410,6 +411,10 @@ def _on_database_created(self, event: DatabaseCreatedEvent) -> None:
             self.unit.status = WaitingStatus("Waiting to connect to Hydra container")
             return
 
+        if not self._validate_config_log_level():
+            event.defer()
+            return
+
```
natalian98, line 417: could you add a unit test for it?

bencekov: I took this out, for log_level the only relevant event handlers are pebble_ready and config_change.

src/charm.py
```diff
@@ -296,6 +351,9 @@ def _handle_status_update_config(self, event: HookEvent) -> None:
             self.unit.status = WaitingStatus("Waiting to connect to Hydra container")
             return
 
+        if not self._validate_config_log_level():
+            return
+
```
natalian98, line 356: Let's move the check before calling `template.render`

natalian98 (dismissed):

gruyaume (dismissed):

natalian98 (approved):
lgtm, but please wait for the CI to pass

PR #69, opened 2023-05-24
Convert ingress provided URLs to https

nsklikas:
Convert ingress provided URLs to https and remove the port

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -10,6 +10,7 @@ import json
 import logging
 from os.path import join
 from typing import Any, Dict, Optional
+from urllib.parse import urlparse, urlunparse
 
 from charms.data_platform_libs.v0.data_interfaces import (
     DatabaseCreatedEvent,
@@ -60,6 +61,7 @@ from ops.model import (
 from ops.pebble import ChangeError, Error, ExecError, Layer
 
 from hydra_cli import HydraCLI
+from utils import normalise_url, remove_none_values
 
 logger = logging.getLogger(__name__)
 
@@ -69,11 +71,6 @@ HYDRA_PUBLIC_PORT = 4444
 SUPPORTED_SCOPES = ["openid", "profile", "email", "phone"]
 PEER = "hydra"
 LOG_LEVELS = ["panic", "fatal", "error", "warn", "info", "debug", "trace"]
-
-
-def remove_none_values(dic: Dict) -> Dict:
-    """Remove all entries in a dict with `None` values."""
-    return {k: v for k, v in dic.items() if v is not None}
 
 
 class HydraCharm(CharmBase):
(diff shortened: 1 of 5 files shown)


src/utils.py
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python3
```
nsklikas, line 1: Any objections to adding a `utils.py` file? cc @natalian98 @gruyaume

src/utils.py
```diff
@@ -0,0 +1,36 @@
...
+from typing import Dict
+from urllib.parse import urlparse, urlunparse
+
+
+def remove_none_values(dic: Dict) -> Dict:
+    """Remove all entries in a dict with `None` values."""
+    return {k: v for k, v in dic.items() if v is not None}
+
+
+def normalise_url(url: str) -> str:
```
nsklikas, line 16: This function will be used by both Kratos and Hydra. But since it is a temporary hack until traefik solves this problem for us, I decided not to put it in a library and just duplicate the code.

src/utils.py
```diff
@@ -0,0 +1,42 @@
...
+        http://ingress:80/path/subpath -> https://ingress/path/subpath
+
+    This conversion works under the following assumptions:
+    1) The ingress will serve https under the 443 port, the user-agent will
+       implicitly make the request on that port
+    2) The provided URL is not a relative path
+    3) No user/password is provided in the netloc
+
+    This is a hack and should be removed once traefik provides a way for us to
+    request the https URL.
```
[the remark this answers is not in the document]
nsklikas: Currently there is no bug for this, I had a discussion with Ryan some time ago about it. But I am not sure what the state currently is.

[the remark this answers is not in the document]
natalian98: It might be good to open a bug, otherwise let's have a jira card to revisit it when it's solved.

src/charm.py
```diff
@@ -211,6 +207,24 @@ def _hydra_service_is_running(self) -> bool:
             return False
         return service.is_running()
 
+    @property
+    def _public_url(self):
+        # TODO: The else part does not make much sense, remove it
```
natalian98, line 212: Any idea what could we do instead? We could defer rendering the conf file until the urls are ready, but I find it error prone.

tests/unit/test_utils.py
natalian98, line 1: You could add return typing, as we have it in other tests

src/charm.py
nsklikas: Not sure tbh. This falls under the whole discussion of status handling. I think it would make most sense to return `None` in the `else` part, which would cause the charm to block unless some `dev` flag is set which would enable us to test it without having to setup the whole bundle.

natalian98: @gruyaume what do you think?

gruyaume: - BlockedStatus makes sense if we're waiting for someone to take action. In this case, we're waiting on the ingress charm (traefik) to provide the URL's right? So I think having the charm in WaitingStatus makes more sense.
- I don't know about deferring. Does the lib for traefik has a custom event for informing that the url is available? If yes no need to defer, let's only leverage this event.

nsklikas: WaitingStatus makes sense when the relation is there, but the url is not ready. 
If there is no relation (ie the user has not connected the 2 units) I think that it makes to block then, because they need to perform an action.

gruyaume: Agreed

bencekov (approved):
Looks good!

gruyaume (approved):

PR #80, opened 2023-06-16
Generate secrets and store them in juju secrets

nsklikas:
Generate cookie and system secrets and store them in juju secrets. 

Cookie secrets are used for signing cookies and system secrets are used for encrypting data stored in the database. The user may want access to the system secrets after the hydra application has been removed from juju as they may wish to access the data stored in the database, currently there is no solution for this use case.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -430,7 +430,7 @@ class HydraCharm(CharmBase):
             return
 
         if not self._get_secrets():
-            self.unit.status = WaitingStatus("Waiting for secret creation")
+            self.unit.status = WaitingStatus("Waiting for secrets creation")
             event.defer()
             return
 


src/charm.py
```diff
@@ -394,10 +420,22 @@ def _handle_status_update_config(self, event: HookEvent) -> None:
             self.unit.status = WaitingStatus("Waiting for database creation")
             return
 
+        if not self._get_secrets():
+            self.unit.status = WaitingStatus("Waiting for secret creation")
```
natalian98, line 424:
```suggestion
            self.unit.status = WaitingStatus("Waiting for secrets creation")
```

src/charm.py
```diff
@@ -356,6 +367,21 @@ def _pop_oauth_relation_peer_data(self, relation_id: int) -> Dict:
...
+            return juju_secret.get_content()
+        except SecretNotFoundError:
+            return None
+
+    def _create_secrets(self) -> Optional[Secret]:
+        if not self.unit.is_leader():
+            return None
+
+        secret = {COOKIE_SECRET_KEY: token_hex(16), SYSTEM_SECRET_KEY: token_hex(16)}
+        juju_secret = self.model.app.add_secret(secret, label=SECRETS_LABEL)
```
natalian98, line 382: Is it ok to have multiple keys in one juju secret? @gruyaume

natalian98 (commented):
Looks good to me, I'd just like to make sure that it's fine to have multiple keys in one juju secret.

src/charm.py
nsklikas: I think that since they are for the same use, which means that they will most of the time be fetched together and created together, it is fine. 

Now that I think about it though, if rotation is implemented we may want to separate them, so as to allow them to be rotated with different frequency. I think I should split them to different secrets after all.

nsklikas: I separated them.

natalian98 (approved):

natalian98 (dismissed):

natalian98 (approved):

Issue #81, opened 2023-06-19
Have the oauth lib observe relation-broken on behalf of the charm


sed-i:
Observe a custom event instead of a relation event.

Similar to `self.oauth.on.oauth_info_changed`, it would be nice to abstract/hide the particular relation event behind a custom event from the lib.

- Perhaps relation-broken should emit the same `oauth_info_changed` custom event?
- If relation-broken has a different semantic then perhaps add a new one, e.g. `on.oath_removed`?

This way you'd potentially be able to hide implementation detail. (What if in the future you decide to emit the custom event on relation-departed instead?)

_Originally posted by @sed-i in https://github.com/canonical/grafana-k8s-operator/pull/213#discussion_r1191453989_

PR #83, opened 2023-06-28
Add custom event for oauth `relation-broken` and remove provider data on `relation-departed`

natalian98:
This PR:
- removes the provider relation data on `relation-departed`
- adds custom `on_oauth_info_removed` event to handle `relation-broken` in order to notify the requirer charm that the relation data was removed

closes https://github.com/canonical/hydra-operator/issues/81

### DIFF
--- a/lib/charms/hydra/v0/oauth.py
+++ b/lib/charms/hydra/v0/oauth.py
@@ -56,7 +56,13 @@ from dataclasses import asdict, dataclass, field
 from typing import Dict, List, Mapping, Optional
 
 import jsonschema
-from ops.charm import CharmBase, RelationChangedEvent, RelationCreatedEvent, RelationDepartedEvent
+from ops.charm import (
+    CharmBase,
+    RelationBrokenEvent,
+    RelationChangedEvent,
+    RelationCreatedEvent,
+    RelationDepartedEvent,
+)
 from ops.framework import EventBase, EventSource, Handle, Object, ObjectEvents
 from ops.model import Relation, Secret, TooManyRelatedAppsError
 
@@ -68,7 +74,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 4
+LIBPATCH = 5
 
 logger = logging.getLogger(__name__)
 
@@ -310,10 +316,23 @@ class InvalidClientConfigEvent(EventBase):
         self.error = snapshot["error"]
 
 
+class OAuthInfoRemovedEvent(EventBase):
+    """Event to notify the charm that the client data was removed."""
+
+    def snapshot(self) -> Dict:
+        """Save event."""
+        return {}
+
+    def restore(self, snapshot: Dict) -> None:
+        """Restore event."""
+        pass
+
+
 class OAuthRequirerEvents(ObjectEvents):
     """Event descriptor for events raised by `OAuthRequirerEvents`."""
 
     oauth_info_changed = EventSource(OAuthInfoChangedEvent)
+    oauth_info_removed = EventSource(OAuthInfoRemovedEvent)
     invalid_client_config = EventSource(InvalidClientConfigEvent)
 
 
@@ -335,12 +354,22 @@ class OAuthRequirer(Object):
         events = self._charm.on[relation_name]
         self.framework.observe(events.relation_created, self._on_relation_created_event)
         self.framework.observe(events.relation_changed, self._on_relation_changed_event)
+        self.framework.observe(events.relation_broken, self._on_relation_broken_event)
 
     def _on_relation_created_event(self, event: RelationCreatedEvent) -> None:
         try:
             self._update_relation_data(self._client_config, event.relation.id)
         except ClientConfigError as e:
             self.on.invalid_client_config.emit(e.args[0])
+
+    def _on_relation_broken_event(self, event: RelationBrokenEvent) -> None:
+        if self.is_client_created():
+            event.defer()
+            logger.info("Relation data still available. Deferring the event")
+            return
+
+        # Notify the requirer that the relation data was removed
+        self.on.oauth_info_removed.emit()
 
     def _on_relation_changed_event(self, event: RelationChangedEvent) -> None:
         if not self.model.unit.is_leader():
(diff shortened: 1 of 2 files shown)


lib/charms/hydra/v0/oauth.py
```diff
@@ -401,8 +430,8 @@ def is_client_created(self, relation_id: Optional[int] = None) -> bool:
             return None
 
         return (
-            "client_id" in relation.data[relation.app]
-            and "client_secret_id" in relation.data[relation.app]
+            relation.data[relation.app]["client_id"]
+            and relation.data[relation.app]["client_secret_id"]
```
nsklikas, line 434: How do we know that these keys are in the relation and will not throw a `KeyError`?

lib/charms/hydra/v0/oauth.py
```diff
@@ -662,6 +694,17 @@ def _delete_juju_secret(self, relation: Relation) -> None:
         secret = self.model.get_secret(label=self._get_secret_label(relation))
         secret.remove_all_revisions()
 
+    def _pop_client_data(self, relation_id: Relation) -> None:
+        if len(self.model.relations) == 0:
+            return None
```
nsklikas, line 699: The convention is to not specify the None if the return type is `None` (ie not optional)
```suggestion
            return
```

lib/charms/hydra/v0/oauth.py
```diff
@@ -662,6 +694,17 @@ def _delete_juju_secret(self, relation: Relation) -> None:
...
+
+        relation = self.model.get_relation(self._relation_name, relation_id=relation_id)
+        if not relation or not relation.app:
+            return None
```
nsklikas, line 703: The convention is to not specify the None if the return type is `None` (ie not optional)
```suggestion
            return
```

lib/charms/hydra/v0/oauth.py
```diff
@@ -662,6 +694,17 @@ def _delete_juju_secret(self, relation: Relation) -> None:
...
+
+        for provider_data in list(relation.data[self.model.app]):
+            relation.data[self.model.app].pop(provider_data, "")
```
nsklikas, line 706: I believe that this can fail. The function body is not an atomic operation which means that the data may be gone when this runs (maybe the data was never there). I think we need to wrap this in a try/catch.

lib/charms/hydra/v0/oauth.py
```diff
@@ -335,13 +354,23 @@ def __init__(
...
             self._update_relation_data(self._client_config, event.relation.id)
         except ClientConfigError as e:
             self.on.invalid_client_config.emit(e.args[0])
 
+    def _on_relation_broken_event(self, event: RelationBrokenEvent) -> None:
+        if self.is_client_created():
+            event.defer()
+            logger.info("Relation data still available. Deferring the event")
+            return
+
```
nsklikas, line 370: Why don't we pop the relation data like in the provider side?

nsklikas (changes requested):
It feels like in this PR we have 2 different solutions to the same problem: do we defer the event or do we manually delete the data? We should use only one of them (I don't really have a preference as both seem hacky, the one with popping the data is probably more correct)

lib/charms/hydra/v0/oauth.py
natalian98: The relation data is still present on relation -departed and -broken. If it was gone (like you'd expect from the relation docs), this method would not be needed at all. If `relation.data[self.model.app]` was empty for whichever reason, this method would just have nothing to pop.
That said, I think it doesn't hurt to wrap it in a try/catch, so I'll update the PR.

lib/charms/hydra/v0/oauth.py
natalian98: I focused on removing the provider's data, but that's a good point, I'll pop the requirer too.

natalian98:
> It feels like in this PR we have 2 different solutions to the same problem: do we defer the event or do we manually delete the data? We should use only one of them (I don't really have a preference as both seem hacky, the one with popping the data is probably more correct)

In this PR I'm trying to address 2 issues that I believe are related to each other:
1. The juju bug with the relation data still being accessible by charm code while inside the `relation-broken` hook.
- As a workaround to invalidate the relation data, I pop it on `relation-departed`.

2. A custom `oauth-info-removed` event.
Apart from [this](https://github.com/canonical/hydra-operator/issues/81) request to hide the `relation-broken` event behind a custom event in the oauth lib, there is a need for it due to the following issue:
- when you `juju remove-relation`, both the requirer (e.g. Grafana) and the provider (Hydra) get the `relation-departed` -> `relation-broken` events fired
- Those events may be launched quicker for one of the relation ends. So for example, the requirer may have its `relation-departed` hook run before the provider. In that case, the requirer still sees the data provided by Hydra, because Hydra didn't manage to pop it "on time"
- To give a specific example, grafana (requirer) needs to restart its service on relation-departed. Its charm logic adds some extra generic oauth variables based on `is_client_created` check. That check is expected to return false on relation departed/broken, because the client data should not be in the databag at that point.
Now, due to the databag presence, `is_client_created=True` and grafana proceeds to add the extra env variables.
- Because of the above, the requirer's `relation-broken` event needs to be deferred until it makes sure that hydra removed its relation data. Once the data is removed, a custom `oauth_info_removed` event is emitted.
- When that happens, the requirer can re-configure itself as needed.

lib/charms/hydra/v0/oauth.py
```diff
@@ -199,6 +205,24 @@ def _dump_data(data: Dict, schema: Optional[Dict] = None) -> Dict:
     return ret
 
 
+class OAuthRelation(Object):
+    """A class containing helper methods for oauth relation."""
+
+    def _pop_relation_data(self, relation_id: Relation) -> None:
```
nsklikas, line 211: Only the leader should run this

nsklikas (commented):
Other than a minor comment LGTM. 

I have a question, if the `relation_broken` event is fired and you pop the relation data, does that trigger a `relation_changed` event on the other side of the relation? From what I gather the `relation.app` is not None, but it is an `Application` with empty name (not sure if that is still true), could we check for that instead of popping the data?

I have to say that this looks stupid (not your fault)

natalian98:
> I have a question, if the `relation_broken` event is fired and you pop the relation data, does that trigger a `relation_changed` event on the other side of the relation? From what I gather the `relation.app` is not None, but it is an `Application` with empty name (not sure if that is still true), could we check for that instead of popping the data?

I checked it, the `relation.app` is an `ops.model.Application provider` object, like on other events.
Popping the requirer data doesn't trigger the `oauth-relation-changed` event after the relation departed.


> I have to say that this looks stupid (not your fault)

:clown_face:

nsklikas (approved):

PR #90, opened 2023-07-13
Hydra alert rules and dashboard

bencekov:
# TESTING

To test Hydra injecting alert rules and dashboards you will need the cos-lite bundle, and postgresql:
```
juju deploy cos-lite --trust
juju deploy postgresql-k8s --trust
```

Pack and deploy hydra (in hydra operator root directory):

```
charmcraft pack
juju deploy ./hydra*.charm  --resource oci-image=$(yq eval '.resources.oci-image.upstream-source' metadata.yaml) --trust
juju integrate hydra postgresql-k8s
```

# Integrations:

You need to integrate hydra with prometheus, loki, and grafana:

```
juju integrate hydra prometheus
juju integrate hydra loki
juju integrate hydra:grafana:dashboard grafana
```

After this you may access grafana's UI with your browser. The url is grafana-ip:3000. Username is admin, you can get the password with `juju run grafana/0 get-admin-password`

On the grafana interface you should be able to find the dashboard, and the alerting rules.

Currently the alert rules have very low comparison values in their expressions. This is to be easier to trigger them, thus verifying that they work. The final values are not set yet.


src/prometheus_alert_rules/hydra_missing.rule
```diff
@@ -0,0 +1,8 @@
+alert: HydraMissing
+expr: up == 0
+for: 1m
+labels:
+  severity: critical
+annotations:
+  summary: Prometheus target missing (instance {{ $labels.instance }})
+  description: "Hydra target has disappeared. An exporter might be crashed.\n  VALUE = {{ $value }}\n  LABELS = {{ $labels }}"
```
wood-push-melon, line 8: I am a little bit confused about this rule with the `HydraUnitDown`. Do they both infer to hydra is not healthy with the only difference in healthy unit number? If so, I guess we could just provide one rule to cover, e.g. the ratio of healthy units is below a certain threshold for duration x.

src/loki_alert_rules/high_error_rate.rule
```diff
@@ -0,0 +1,8 @@
+alert: HostHighLogErrorRate
+expr: count_over_time({%%juju_topology%%} |= "error" [1m]) > 0
+for: 1m
+labels:
+  severity: error
+annotations:
+  summary: High error rate in logs (instance {{ $labels.instance }})
```
wood-push-melon, line 7: Is this relatively duplicated with the above rules? From the description, it seems to me that they are all trying to detect something based on error logs.

Also, not very sure if the `groups` are allowed here. If it is allowed, I guess it would be better if we could group similar kinds of rules in a `group`.

src/loki_alert_rules/error_level_logs_found.rule
```diff
@@ -0,0 +1,7 @@
+alert: ErrorLevelLogsFound
```
wood-push-melon, line 1: IMHO, I think we could combine the error-level logs with the following fatal-level logs, and panic-level logs together. From a user perspective, those metrics are all important to monitor. Users definitely should act no matter what the log level is.

src/loki_alert_rules/database_unavailable.rule
```diff
@@ -0,0 +1,7 @@
+alert: DatabaseUnavailable
+expr: rate({%%juju_topology%%} |~ `failed to connect to.*host=postgresql-k8s`[1m]) > 0
```
wood-push-melon, line 2: I am new to PromQL. I do have a question about this query. Does it rely on this pattern `failed to connect to blah blah` to detect the error logs? If this is the case, I would not do it this way personally.

And another thing is that this metric does not necessarily shows the database is unavailable. For example, the network issue could also lead to connection issues. IMHO, I would not provide this rule here personally. I think it is more appropriate for the database itself to provide such metrics, e.g. the database could have some alert rule saying how many database units are down, etc.

src/grafana_dashboards/hydra.json.tmpl
```diff
@@ -0,0 +1,693 @@
+{
```
shipperizer, line 1: so there are a chunk of variables that we are not using, mainly the ones `juju` related, i'd say to scrap them as we dont benefit from them (right now at least)

src/grafana_dashboards/hydra.json.tmpl
```diff
@@ -0,0 +1,693 @@
...
+      },
+      "pluginVersion": "9.2.1",
+      "targets": [
+        {
+          "datasource": {
+            "type": "loki",
+            "uid": "${LokiDS}"
+          },
+          "editorMode": "code",
+          "expr": "sum by (juju_unit) (count_over_time( {juju_charm=\"hydra\"} |= `\"level\":\"error\"` |= `\"audience\":\"application\"`[5m]))",
```
shipperizer, line 628:
```suggestion
          "expr": "sum by (juju_unit) (count_over_time( {juju_charm=\"hydra\"} |= `\"level\":\"error\"` |= `\"audience\":\"application\"`[$__interval]))",
```

shipperizer: logs query needs to be time dependent

src/grafana_dashboards/hydra.json.tmpl
```diff
@@ -0,0 +1,693 @@
...
+        }
+      },
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${PrometheusDS}"
+          },
+          "editorMode": "code",
+          "expr": "sum(http_requests_statuses_total) by (status_bucket, juju_unit)",
```
shipperizer, line 314: i'd scrap this, as the value tends to increase and then becomes useless, the `increase` is a better one to monitor status of the apps

src/grafana_dashboards/hydra.json.tmpl
```diff
@@ -0,0 +1,693 @@
...
+        }
+      },
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${PrometheusDS}"
+          },
+          "editorMode": "code",
+          "expr": "histogram_quantile(0.9, sum(rate(http_requests_duration_seconds_bucket{juju_charm=\"hydra\"}[5m])) by (le, juju_unit))",
```
shipperizer, line 406:
```suggestion
          "expr": "histogram_quantile(0.9, sum(rate(http_requests_duration_seconds_bucket{juju_charm=\"hydra\"}[5m])) by (le))",
```

shipperizer: in general we don't want a drilldown per pod/app for health metrics, basically we wanna know how many units/pods/process are healthy out of the total we have deployed, but for `p90` (which is this query here) we care about the aggregate, basically we wanna go the full hog and not have a separate view per pod (which is only useful when debugging)

src/grafana_dashboards/hydra.json.tmpl
```diff
@@ -0,0 +1,693 @@
...
+      },
+      "pluginVersion": "9.2.1",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${PrometheusDS}"
+          },
+          "editorMode": "code",
+          "expr": "sum(up{juju_charm=\"hydra\"}) by (juju_unit)",
```
shipperizer, line 564: as per se, this doesn't tell us how many pods are there vs how many are healthy, we might need extra metrics to combine

also noticed we don't have the generic `kube_pod` metrics (at least can't see them in my microk8s) which would be good in this case to use as comparison, but allegedly combining the above with the `kube_pod_status_running` would give us the availability

src/loki_alert_rules/database_unavailable.rule
shipperizer: i agree with @wood-push-melon

src/loki_alert_rules/error_level_logs_found.rule
shipperizer: same here, i totally agree with @wood-push-melon

src/loki_alert_rules/high_error_rate.rule
shipperizer: :+1:

src/prometheus_alert_rules/hydra_missing.rule
shipperizer: comparison number should be the amount of hydra pods running, also this will allegedly fire for any component monitored by prometheus

src/prometheus_alert_rules/high_restart_rate.rule
```diff
@@ -0,0 +1,8 @@
+alert: HydraHighRestartRate
+expr: rate(process_start_time_seconds[3m]) > 0.1
```
shipperizer, line 2: which magic adds the labels so that this query (even though the metric is hydra specific here) doesn't trigger on other applications?as i see that `%%juju_topology%%` is missing

src/prometheus_alert_rules/hydra_slow_request_handling.rule
```diff
@@ -0,0 +1,8 @@
+alert: HydraSlowRequestHandling
+expr: histogram_quantile(0.9, sum(rate(http_requests_duration_seconds_bucket[5m])) by (le)) > 0.4
```
shipperizer, line 2: as much as this is a good metric to chuck in here, we have no clue about response profiling so not sure how valuable an alert would be when the p90 is set to a random number

src/prometheus_alert_rules/high_restart_rate.rule
bencekov: For prometheus rules, the labels are automatically filled in to the template. The %%juju_topology%% is something you should only see on loki rules.

src/grafana_dashboards/hydra.json.tmpl
bencekov: just to clarify, by increase you mean the query with the sum(rate(http_statuses))?

src/loki_alert_rules/database_unavailable.rule
bencekov: I should have phrased it better, but the alert exists to notify if the hydra can not access the backend, not to be a replacement for alerts for the postgresql backend.

PR #92, opened 2023-07-24
IAM-378-Add alert rules to Hydra operator

bencekov:
Added alert rules to Hydra

# TESTING

To test Hydra alerts, you will need the cos-lite bundle, and the data backend:
```
juju deploy cos-lite --trust
juju deploy postgresql-k8s --trust
```

Pack and deploy Hydra (in operator root directory):

```
charmcraft pack
juju deploy ./hydra*.charm  --resource oci-image=$(yq eval '.resources.oci-image.upstream-source' metadata.yaml) --trust
juju integrate hydra postgresql-k8s
```

# Integrations:

You need to integrate the hydra operator with prometheus, loki:

```
juju integrate hydra prometheus
juju integrate hydra loki
```

You can verify the alerts on the Grafana's UI interface. After this you may access Grafana's UI with your browser. The url is grafana-ip:3000. Username is admin, you can get the password with `juju run grafana/0 get-admin-password`

On the Grafana interface you should be able to find options bar on the left side of the screen. Go to alerts. There you will see the available alerts from prometheus and loki. You can open them up with the arrow on the left to see more details, or to open a graph mapping the result of the alert query over time.

### DIFF
--- a/src/loki_alert_rules/hydra_high_severity_log.rule
+++ b/src/loki_alert_rules/hydra_high_severity_log.rule
@@ -2,13 +2,13 @@ groups:
 - name: HydraHighSeverityLog
   rules:
   - alert: LowFrequencyHighSeverityLog
-    expr: sum by(severity) (count_over_time({%%juju_topology%%} | json | severity =~ `error|fatal|critical` [5m])) > 0 and sum by(severity) (count_over_time({%%juju_topology%%} | json | severity =~ `error|fatal|critical` [5m])) < 100
+    expr: sum by(level) (count_over_time({%%juju_topology%%} | json | level =~ `error|fatal|critical` [5m])) > 0 and sum by(level) (count_over_time({%%juju_topology%%} | json | level =~ `error|fatal|critical` [5m])) < 100
     labels:
       severity: warning
     annotations:
       summary: "Logs with level error or above found in application {{ $labels.juju_application }} of Juju charm {{ $labels.juju_charm }} in model {{ $labels.juju_model }}. Frequency of logs is low."
   - alert: HighFrequencyHighSeverityLog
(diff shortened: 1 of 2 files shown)


src/loki_alert_rules/hydra_high_severity_log.rule
```diff
@@ -0,0 +1,39 @@
+groups:
+- name: HydraHighSeverityLog
+  rules:
+  - alert: ErrorLevelLog
+    expr: sum(count_over_time({%%juju_topology%%} |= `"level":"error"` [5m])) < 100 and sum(count_over_time({%%juju_topology%%} |= `"level":"error"` [5m])) > 0
+    labels:
+      severity: warning
+    annotations:
+      message: "Logs with level error found in application {{ $labels.juju_application }} of Juju charm {{ $labels.juju_charm }} in model {{ $labels.juju_model }}. Frequency of logs with level error is low."
```
shipperizer, line 9: i think u can swap the multiple log queries with this one 

```
sum by(level) (count_over_time({juju_application="hydra"} | json | level =~ `error|fatal|critical` [5m]))
```

src/loki_alert_rules/hydra_high_severity_log.rule
```diff
@@ -0,0 +1,39 @@
...
+  - alert: ErrorLevelLogHighFrequency
+    expr: sum(count_over_time({%%juju_topology%%} |= `"level":"error"` [5m])) > 100
```
shipperizer, line 11: same here but only if threshold are the same across different levels

shipperizer (approved):
lgtm

in terms of commit history when you feel that u have loads of small `fix this` and `fix that` where the diff between them is actually not that meaningful you might want to have a look at `git reset --soft <start commit>` and then group them in a more siloed and meaningful way

PR #95, opened 2023-07-28
IAM-389-Add Tracing to Hydra

bencekov:
This pr adds the tracing integration to the Hydra Operator
---
Testing:
The tracing configurations for hydra are in envars.
A unit test is in there to make sure that the correct pebble layer gets rendered after relation with tempo.

# Manual Test
You will need to deploy tempo-k8s, hydra operator, postgresql-k8s, grafana-k8s to verify the that tracing is working.
Integrate hydra with the data backend, then with tempo-k8s.
```
juju integrate hydra tempo-k8s
```
Integrate tempo-k8s with grafana as a grafana source:
```
juju integrate tempo-k8s grafana-k8s:grafana-source
```

Open grafana in the browser: grafana_ip:3000
login name: admin
to get password run:
```
juju run grafana-k8s/0 get-admin-password
```

Go to the explore button in the menu. Pick tempo as the data source. Set the query type to search. Enter Ory Hydra to Service Name. Start Query. If you see results the integration was succesful.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -35,6 +35,7 @@ from charms.identity_platform_login_ui_operator.v0.login_ui_endpoints import (
 from charms.loki_k8s.v0.loki_push_api import LogProxyConsumer, PromtailDigestError
 from charms.observability_libs.v0.kubernetes_service_patch import KubernetesServicePatch
 from charms.prometheus_k8s.v0.prometheus_scrape import MetricsEndpointProvider
+from charms.tempo_k8s.v0.tracing import TracingEndpointProvider
 from charms.traefik_k8s.v1.ingress import (
     IngressPerAppReadyEvent,
     IngressPerAppRequirer,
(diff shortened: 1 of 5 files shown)


src/charm.py
```diff
@@ -247,6 +256,42 @@ def _hydra_layer(self) -> Layer:
                 },
             },
         }
+
+        layer_with_trace_config = {
```
natalian98, line 260: The only difference between `layer_config` and `layer_with_trace_config` are the environment variables. Can we add that part based on the `self._tracing_ready` check instead of keeping both layers separate?

src/charm.py
```diff
@@ -247,6 +256,42 @@ def _hydra_layer(self) -> Layer:
...
+                "version": {
+                    "override": "replace",
+                    "exec": {"command": "hydra version"},
+                },
+                "ready": {
+                    "override": "replace",
+                    "http": {"url": f"http://localhost:{HYDRA_ADMIN_PORT}/health/ready"},
+                },
+            },
+        }
```
nsklikas, line 291: I think this would be cleaner if you would construct the service object first with conditionals and then construct the layer, something like:
```python
container = {...}
if self._tracing_ready:
    container["environment"] = {...}
layer_config = {
    ...
    "services": {
        self._container_name: container
    }
} 
```

shipperizer:
little update with IAM-394

i'd say we merge this once fixed but then we make another branch off `main` ready for when the new stuff goes in

src/charm.py
bencekov: sure

src/charm.py
bencekov: sure

PR #106, opened 2023-08-18
fix: use http endpoint in endpoint-info integration

wood-push-melon:
This pull request aims to do the following:

- provide public endpoints with HTTP protocol in the `hydra-endpoint-info` integration
- update the endpoint whenever there is a change in the `ingress` integration

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -303,22 +303,14 @@ class HydraCharm(CharmBase):
         self._handle_status_update_config(event)
 
     @property
-    def _public_url(self) -> str:
-        # TODO: The else part does not make much sense, remove it
-        return (
-            normalise_url(self.public_ingress.url)
-            if self.public_ingress.is_ready()
-            else f"http://127.0.0.1:{HYDRA_PUBLIC_PORT}/"
-        )
+    def _public_url(self) -> Optional[str]:
+        url = self.public_ingress.url
+        return normalise_url(url) if url else None
 
     @property
-    def _admin_url(self) -> str:
-        # TODO: The else part does not make much sense, remove it
-        return (
-            normalise_url(self.admin_ingress.url)
-            if self.admin_ingress.is_ready()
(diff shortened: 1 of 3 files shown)


src/charm.py
```diff
@@ -338,7 +330,7 @@ def _render_conf_file(self) -> str:
             consent_url=self._get_login_ui_endpoint_info("consent_url"),
             error_url=self._get_login_ui_endpoint_info("oidc_error_url"),
             login_url=self._get_login_ui_endpoint_info("login_url"),
-            hydra_public_url=self._public_url,
+            hydra_public_url=self._public_url or f"http://127.0.0.1:{HYDRA_PUBLIC_PORT}/",
```
wood-push-melon, line 333: Let me know if I should put the `http://127.0.0.1:4444` into the config file `hydra.yaml`.

bencekov: I think it's better to have it here than in the config file. Is this a necessary change though?

natalian98: The `self: issuer` and [public](https://github.com/canonical/hydra-operator/blob/5292f086517bf7da6206e6212d661330d65225b8/templates/hydra.yaml.j2#L29) configs need to have some value for hydra to run, so it's fine to put `http://127.0.0.1:{HYDRA_PUBLIC_PORT}/` as a fallback in case the traefik public url is not available.

bencekov (approved):

src/charm.py
nsklikas: Doesn't make much of a difference, but since it's required to have it in the `hydra.yaml` then I would argue that it should be placed in the jinja template (else the produced file is not correct).

nsklikas (approved):

src/charm.py
wood-push-melon: thx, I'll give it in another pr.

PR #111, opened 2023-08-31
IAM-460-Update login-ui relation

bencekov:
As part of IAM-460, I took out the unused pages from the login-ui backend as well as from the login_ui_endpoints integration lib. This pr accommodates the relation update in charm.py.


nsklikas:
I don't think that this PR is needed, it will be done automatically by the update-libs workflow

bencekov:
> I don't think that this PR is needed, it will be done automatically by the update-libs workflow

I think the unit tests should reflect the state of the relation.

natalian98 (approved):

PR #134, opened 2023-09-20
Expose app version to juju

nsklikas:
IAM-493

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -345,6 +345,10 @@ class HydraCharm(CharmBase):
         )
         return rendered
 
+    def _set_version(self) -> None:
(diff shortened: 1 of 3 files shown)


nsklikas:
This change will expose the hydra version as seen below:
image

Most of our charmed applications do not have the `v` prefix, should we filter that out?

shipperizer (approved):

shipperizer:
> This change will expose the hydra version as seen below: image
> 
> Most of our charmed applications do not have the `v` prefix, should we filter that out?

is that the version of the charm or the app version?if former i'd say filter that out

PR #138, opened 2023-09-26
Fix database/migration logic

nsklikas:
IAM-513

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -373,6 +373,21 @@ class HydraCharm(CharmBase):
         return f"oauth_{relation_id}"
 
     @property
+    def _migration_peer_data_key(self) -> Optional[str]:
+        if not self.database.relations:
+            return None
+        # We append the relation ID to the migration key in peer data, this is
+        # needed in order to be able to store multiple migration versions.
+        #
+        # When a database relation is departed, we can't remove the key because we
+        # can't be sure if the relation is actually departing or if the unit is
+        # dying. If a new database relation is then created we need to be able to tell
+        # that it is a different relation. By appending the relation ID we overcome this
+        # problem.
+        # See https://github.com/canonical/hydra-operator/pull/138#discussion_r1338409081
+        return f"{DB_MIGRATION_VERSION_KEY}_{self.database.relations[0].id}"
+
+    @property
     def _peers(self) -> Optional[Relation]:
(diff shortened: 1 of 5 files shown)


src/charm.py
```diff
@@ -593,11 +596,21 @@ def _on_run_migration(self, event: ActionEvent) -> None:
...
+    def _cleanup_peer_data(self) -> None:
+        if not self._peers:
+            return
+        # We need to remove the migration key from peer data. We can't do that in relation
+        # departed as we can't tell if the event was triggered from a unit dying of the
+        # relation being actually departed.
+        extra_keys = [
+            k
+            for k in self._peers.data[self.app].keys()
+            if k.startswith(DB_MIGRATION_VERSION_KEY) and k != self._migration_peer_data_key
```
wood-push-melon, line 610: I have a question here: why do we use a dynamically changing `key` in the peer databag? 

I thought the `key` would be constant and the value would be updated.

nsklikas: When hydra is related to a database and you remove the relation, we need to remove the migration information from the peer relation. The problem with this is that AFAIK it is impossible to tell if the relation is actually gone or if the unit is just dying. What we could do is to remove the key after the relation is gone (with a periodic job like this one).

The problem with is what would happen if before you clean up the databag of the migration version, you get a new database relation? You will never clean up the key and the migration will never run because the peer data contain the old migration. Your peer databag will look like this:
```
{
  "migration_version": "v2.1.1"
}
```

By adding the relation ID in the key (making it dynamic), so if you get another database relation before the peer relation databag is cleaned you will still be able to tell that it comes from a different DB and that the migration needs to run. The databag will look like this:
(after you remove the first db relation)
```
{
  "migration_version_3": "v2.1.1"
}
```
(when you integrate with a new db connection)
```
{
  "migration_version_3": "v2.1.1",
  "migration_version_10": "v2.1.1"
}
```
(after the clean up job has run)
```
{
  "migration_version_10": "v2.1.1"
}
```

wood-push-melon: I see what you mean. Now I can better understand the initiative.

But I do have another question. Maybe we can sync after the standup.

wood-push-melon (dismissed):

wood-push-melon (approved):

PR #148, opened 2024-02-27
Update _on_relation_changed_event

kian99:
## Description

Change the _on_relation_changed_event to ensure that all units receive the hook. Currently only leader units on the consuming charm will receive the event which is not correct as non-leader units also need the info.


nsklikas (approved):
Thanks for the fix

PR #154, opened 2024-05-02
fix: charmhub review fixes

shipperizer:


src/charm.py
```diff
@@ -280,22 +277,17 @@ def _hydra_layer(self) -> Layer:
...
-            return False
-
-        try:
-            service = self._container.get_service(self._container_name)
-        except (ModelError, RuntimeError):
-            return False
-        return service.is_running()
+        return (
+            self._hydra_service_is_created
+            and self._container.get_service(self._container_name).is_running
```
nsklikas, line 289:
```suggestion
            and self._container.get_service(self._container_name).is_running()
```

nsklikas (approved):

src/charm.py
```diff
@@ -280,22 +277,17 @@ def _hydra_layer(self) -> Layer:
...
-
-        try:
-            service = self._container.get_service(self._container_name)
-        except (ModelError, RuntimeError):
-            return False
-        return service.is_running()
+        return (
+            self._hydra_service_is_created
+            and self._container.get_service(self._container_name).is_running()
+        )
```
weiiwang01, line 290:
```suggestion
        return (
            self._hydra_service_is_created
            and self._container.get_service(self._container_name)
            and self._container.get_service(self._container_name).is_running()
        )
```
Just in case the service doesn't exist?

nsklikas: I think that this handled by `_hydra_service_is_created` (python evaluates conditions lazily). If the service does not exist I think that `_hydra_service_is_created` will return False (because of https://github.com/canonical/hydra-operator/pull/154/files#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cR282) and then https://github.com/canonical/hydra-operator/pull/154/files#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cR289 will never be evaluated.

No problem with adding it though, for clarity's sake

PR #158, opened 2024-05-21
chore: replace black with ruff as the formatter

wood-push-melon:
This pull request:

- replaces black with ruff as the formatter
- fixes the lint and format issues with ruff
- keeps the `isort` to keep the original behaviors for import sorting
- puts the `codespell` configuration to the `pyproject.toml` file to be consistent with other tools

### DIFF
--- a/lint-requirements.txt
+++ b/lint-requirements.txt
@@ -1,9 +1,3 @@
-black
-flake8==6.1.0
-flake8-docstrings
-flake8-copyright
-flake8-builtins
-pyproject-flake8
-pep8-naming
-isort
 codespell
+isort
+ruff
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,33 +12,59 @@ show_missing = true
 minversion = "6.0"
 log_cli_level = "INFO"
 
-# Formatting tools configuration
-[tool.black]
-line-length = 99
-target-version = ["py38"]
+# Linting and formatting tools configuration
+[tool.codespell]
+# https://github.com/codespell-project/codespell
+skip = ".git,.tox,build,lib,venv,.venv,.mypy_cache,icon.svg,__pycache__"
 
 [tool.isort]
+# Profiles: https://pycqa.github.io/isort/docs/configuration/profiles.html
 line_length = 99
 profile = "black"
 
-# Linting tools configuration
-[tool.flake8]
-max-line-length = 99
-max-doc-length = 99
-max-complexity = 10
-exclude = [".git", "__pycache__", ".tox", "build", "dist", "*.egg_info", "venv"]
-select = ["E", "W", "F", "C", "N", "R", "D", "H"]
-# Ignore W503, E501 because using black creates errors with this
-# Ignore D107 Missing docstring in __init__
-ignore = ["W503", "E501", "D107"]
-# D100, D101, D102, D103: Ignore missing docstrings in tests
-per-file-ignores = ["tests/*:D100,D101,D102,D103,D104"]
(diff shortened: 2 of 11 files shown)


lint-requirements.txt
```diff
@@ -1,9 +1,2 @@
-black
-flake8==6.1.0
-flake8-docstrings
-flake8-copyright
```
natalian98, line 4: Does ruff check for copyright notices? I think we'd need to use `--preview` flag: https://docs.astral.sh/ruff/rules/missing-copyright-notice/

pyproject.toml
```diff
@@ -12,33 +12,50 @@ show_missing = true
...
+    "E501",
+    "N818"
+]
+per-file-ignores = {"tests/*" = ["D100","D101","D102","D103","D104"]}
 
-# Linting tools configuration
-[tool.flake8]
-max-line-length = 99
-max-doc-length = 99
+[too.ruff.lint.flake8-copyright]
```
natalian98, line 50:
```suggestion
[tool.ruff.lint.flake8-copyright]
```

pyproject.toml
```diff
@@ -12,33 +12,50 @@ show_missing = true
...
 line-length = 99
-target-version = ["py38"]
+include = ["pyproject.toml", "src/**/*.py", "tests/**/*.py", "lib/charms/hydra/**/*.py"]
+extend-exclude = ["__pycache__", "*.egg_info"]
+target-version = "py38"
 
-[tool.isort]
-line_length = 99
-profile = "black"
+[too.ruff.lint]
```
natalian98, line 24:
```suggestion
[tool.ruff.lint]
```

lint-requirements.txt
wood-push-melon: It does when `preview` is enabled. Just somehow it keeps complaining about it so I disabled it for now. I am currently taking a look at the `CPY` rule and checking where goes wrong.

src/charm.py
```diff
@@ -58,15 +59,12 @@
...
-    ModelError,
     Relation,
     Secret,
     SecretNotFoundError,
     WaitingStatus,
 )
 from ops.pebble import ChangeError, Error, ExecError, Layer
-
-from hydra_cli import HydraCLI
 from utils import normalise_url, remove_none_values
```
nsklikas, line 68: Import sorting seems to have changed, I think we should keep the current approach where local imports are separately grouped

tox.ini
```diff
@@ -28,30 +28,27 @@ deps =
...
+    ruff format {[vars]all_path}
 
 [testenv:lint]
 description = Check code against coding style standards
 deps =
+    tomli
     -r{toxinidir}/lint-requirements.txt
 commands =
-    codespell {[vars]lib_path}
-    codespell {toxinidir}/ --skip {toxinidir}/.git --skip {toxinidir}/.tox \
```
wood-push-melon, line 48: Moved the configuration of `codespell` to `pyproject.toml` file.

tox.ini
```diff
@@ -28,30 +28,27 @@ deps =
...
 
 [testenv:fmt]
-description = Apply coding style standards to code
+description = Apply coding style standards
 deps =
     -r{toxinidir}/fmt-requirements.txt
 commands =
-    isort {[vars]all_path}
-    black {[vars]all_path}
+    ruff format {[vars]all_path}
```
nsklikas, line 41: Don't we need to run isort here as well?

nsklikas (approved):
lgtm

PR #159, opened 2024-05-22
Add jwt_access_token field to relation databag

nsklikas:
After discussing it with @gtato it looks like we should add a jwt at field in the relation to allow it to be used with other providers other than hydra that support jwt access tokens.

If this is merged, I will update the charm relation interface

### DIFF
--- a/lib/charms/hydra/v0/oauth.py
+++ b/lib/charms/hydra/v0/oauth.py
@@ -67,7 +67,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 8
+LIBPATCH = 9
 
 PYDEPS = ["jsonschema"]
 
@@ -128,6 +128,7 @@ OAUTH_PROVIDER_JSON_SCHEMA = {
         },
         "groups": {"type": "string", "default": None},
         "ca_chain": {"type": "array", "items": {"type": "string"}, "default": []},
+        "jwt_access_token": {"type": "string", "default": "False"},
     },
     "required": [
         "issuer_url",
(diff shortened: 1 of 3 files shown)


lib/charms/hydra/v0/oauth.py
```diff
@@ -201,11 +202,29 @@ def _dump_data(data: Dict, schema: Optional[Dict] = None) -> Dict:
...
             except json.JSONDecodeError as e:
                 raise DataValidationError(f"Failed to encode relation json: {e}")
+        elif isinstance(v, bool):
+            ret[k] = str(v)
         else:
             ret[k] = v
     return ret
 
 
+def strtobool(val: str) -> bool:
```
natalian98, line 212: `ValueError` may be raised instead of returning bool
```suggestion
def strtobool(val: str) -> Optional[bool]:
```

natalian98 (approved):

lib/charms/hydra/v0/oauth.py
```diff
@@ -201,11 +202,29 @@ def _dump_data(data: Dict, schema: Optional[Dict] = None) -> Dict:
...
+    """Convert a string representation of truth to true (1) or false (0).
+
+    True values are 'y', 'yes', 't', 'true', 'on', and '1'; false values
+    are 'n', 'no', 'f', 'false', 'off', and '0'.  Raises ValueError if
+    'val' is anything else.
+    """
+    val = val.lower()
```
BarcoMasile, line 219: suggestion(non-blocking): careful here since you're assuming that `val` is never `None` (or this `.lower()` would raise an error), even though I've seen you perform checks outside of this func. Could be better to handle None inside the func mainly to avoid misuse/surprises

BarcoMasile (dismissed):

lib/charms/hydra/v0/oauth.py
nsklikas: Fair

lib/charms/hydra/v0/oauth.py
```diff
@@ -201,11 +202,32 @@ def _dump_data(data: Dict, schema: Optional[Dict] = None) -> Dict:
...
 
 
+def strtobool(val: str) -> bool:
+    """Convert a string representation of truth to true (1) or false (0).
+
+    True values are 'y', 'yes', 't', 'true', 'on', and '1'; false values
+    are 'n', 'no', 'f', 'false', 'off', and '0'.  Raises ValueError if
+    'val' is anything else.
+    """
+    if not val:
```
wood-push-melon, line 219: just a nonblocking question: why do we need this check at the beginning? can we move it to the falsy value check block blow?

lib/charms/hydra/v0/oauth.py
```diff
@@ -201,11 +202,32 @@ def _dump_data(data: Dict, schema: Optional[Dict] = None) -> Dict:
...
             except json.JSONDecodeError as e:
                 raise DataValidationError(f"Failed to encode relation json: {e}")
+        elif isinstance(v, bool):
+            ret[k] = str(v)
         else:
             ret[k] = v
     return ret
 
 
+def strtobool(val: str) -> bool:
```
wood-push-melon, line 212: Another **nonblocking** question: I can see this function is very similar to the one in `distutils.util`. Is it possible that we could use that instead?

nsklikas: It is taken from `distutils`, `distutils` was marked as deprecated on python `3.10` and `3.11` and it was removed on `3.12` (https://peps.python.org/pep-0632/)

lib/charms/hydra/v0/oauth.py
nsklikas: as a guard for `val.lower()`, i changed the check to make it more clear

BarcoMasile (approved):

PR #161, opened 2024-06-27
IAM-943

shipperizer:
IAM-943: hydra internal networking

- **feat: introduce internal ingress**
- **fix: use internal ingress if set, otherwise stick with k8s networking**
- **fix: patch if secret not found**

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -41,6 +41,7 @@ from charms.traefik_k8s.v2.ingress import (
     IngressPerAppRequirer,
     IngressPerAppRevokedEvent,
 )
+from charms.traefik_route_k8s.v0.traefik_route import TraefikRouteRequirer
 from jinja2 import Template
 from ops.charm import (
     ActionEvent,
@@ -51,6 +52,7 @@ from ops.charm import (
     RelationCreatedEvent,
     RelationDepartedEvent,
     RelationEvent,
+    RelationJoinedEvent,
     WorkloadEvent,
 )
 from ops.main import main
@@ -65,6 +67,7 @@ from ops.model import (
 )
 from ops.pebble import ChangeError, Error, ExecError, Layer
 
+from constants import INTERNAL_INGRESS_RELATION_NAME
 from hydra_cli import HydraCLI
 from utils import normalise_url, remove_none_values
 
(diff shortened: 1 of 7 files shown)


src/charm.py
```diff
@@ -528,19 +636,14 @@ def _update_hydra_endpoints_relation_data(self, event: RelationEvent) -> None:
         logger.info("Sending endpoints info")
 
         admin_endpoint = (
-            self._admin_url
+            self._internal_url
             or f"http://{self.app.name}.{self.model.name}.svc.cluster.local:{HYDRA_ADMIN_PORT}"
         )
```
nsklikas, line 641: The `/admin` part is not needed?

src/charm.py
```diff
@@ -1009,6 +1117,32 @@ def _get_tracing_endpoint_info(self) -> str:
...
+
+        Args:
+            event: a :class:`HookEvent` to signal a change we may need to respond to.
+        """
+        if not self.unit.is_leader():
+            return
+
+        # If it's a RelationJoinedEvent, set it in the ingress object
+        if isinstance(event, RelationJoinedEvent):
+            self.internal_ingress._relation = event.relation
```
nsklikas, line 1136: Why is this needed?

src/charm.py
```diff
@@ -228,6 +241,14 @@ def __init__(self, *args: Any) -> None:
             self.loki_consumer.on.promtail_digest_error,
             self._promtail_error,
         )
+        self.framework.observe(
+            self.on[INTERNAL_INGRESS_RELATION_NAME].relation_joined, self._configure_ingress
+        )
+        self.framework.observe(
+            self.on[INTERNAL_INGRESS_RELATION_NAME].relation_changed, self._configure_ingress
+        )
```
nsklikas, line 249: Shouldn't we handle relation revoked as well?

nsklikas (commented):
lgtm overall, some minor questions

src/charm.py
shipperizer: not sure to be fair, implementation is matching grafana's one

happy to remove and try to see if it works

src/charm.py
shipperizer: nope, the /admin is added by client sdks

if we do it then it is like adding and extra one when calling the admin API so it would end up being `/admin/admin/clients`

src/charm.py
shipperizer: chatted and all hell breaks loose if removed

nsklikas (approved):

PR #165, opened 2024-07-22
refactor: refactor command line

wood-push-melon:


shipperizer:
@wood-push-melon we should drop the `chore: merge main` commits and do a proper rebase

src/secret.py
```diff
@@ -0,0 +1,68 @@
...
+from configs import ServiceConfigs
+from constants import (
+    COOKIE_SECRET_KEY,
+    COOKIE_SECRET_LABEL,
+    SYSTEM_SECRET_KEY,
+    SYSTEM_SECRET_LABEL,
+)
+
+
+class Secrets:
```
wood-push-melon, line 17: The `Secrets` class manages all of the juju secrets created and maintained by the charm itself.

src/cli.py
```diff
@@ -0,0 +1,339 @@
...
+
+from constants import ADMIN_PORT, CONFIG_FILE_NAME, DEFAULT_OAUTH_SCOPES, DEFAULT_RESPONSE_TYPES
+from exceptions import MigrationError
+
+logger = logging.getLogger(__name__)
+
+VERSION_REGEX = re.compile(r"Version:\s+(?P<version>v\d+\.\d+\.\d+)")
+
+
+class OAuthClient(BaseModel):
```
wood-push-melon, line 22: The `OAuthClient` acts as the data interface between caller and oauth-client-related cli commands.

templates/hydra.yaml.j2
```diff
@@ -48,3 +51,5 @@ webfinger:
       {%- if access_token_strategy != "opaque" %}
       - hydra.jwt.access-token
       {%- endif %}
+
+dev: {{ dev_mode | d(false) }}
```
wood-push-melon, line 55: The `dev` configuration is moved from the pebble layer to the config file.

templates/ingress.json.j2
```diff
@@ -0,0 +1,80 @@
+{% set identifier = "{}-{}".format(model, app) %}
```
wood-push-melon, line 1: This template holds the internal ingress configuration.

src/charm.py
```diff
@@ -42,126 +35,126 @@
...
-        self._log_path = self._log_dir / "hydra.log"
-
-        self._hydra_cli = HydraCLI(
-            f"http://localhost:{HYDRA_ADMIN_PORT}", self._container, self._hydra_config_path
-        )
+        self.peer_data = PeerData(self.model)
+        self.secrets = Secrets(self.model)
+        self.charm_config = CharmConfig(self.config)
 
-        self.service_patcher = KubernetesServicePatch(
```
wood-push-melon, line 113: This has been replaced by `open_port`

requirements.txt
```diff
@@ -1,7 +1,6 @@
...
-ops>=2.0.0
-lightkube
-lightkube-models
+dacite
 jinja2
-jsonschema
-pydantic<2.0
+ops>=2.0.0
+pydantic~=2.8.2
+yarl
```
wood-push-melon, line 6: This is for handling URLs operations.

natalian98 (dismissed):
Looks good to me.

Could we have a test for upgrading the charm? See [example](https://github.com/canonical/grafana-k8s-operator/blob/main/tests/integration/test_upgrade_charm.py)

wood-push-melon:
> Looks good to me.
> 
> Could we have a test for upgrading the charm? See [example](https://github.com/canonical/grafana-k8s-operator/blob/main/tests/integration/test_upgrade_charm.py)

Thanks. Will add the test for the charm upgrade.

natalian98 (approved):

requirements.txt
BarcoMasile: Do we really need this? It only provides syntactic sugar for URL operations, nothing the std lib cannot handle.
I think we can live without it and not bloat the project.

If I'm missing something and you we need to keep it, you could remove it from here and put it in `integration-requirements.txt`, I see it is only used in integration tests. Correct me if I'm wrong

wood-push-melon: I think it's also used in the source code, not only in the tests. The stdlib cannot take care of the relative path as we want in the charm. (see [this](https://stackoverflow.com/questions/10893374/python-confusions-with-urljoin/10893427#10893427))

shipperizer (approved):

PR #166, opened 2024-08-01
Migrate to charmcraft.yaml

nsklikas:
IAM-961


charmcraft.yaml
```diff
@@ -10,4 +79,7 @@ bases:
       - name: "ubuntu"
         channel: "22.04"
 
-
+parts:
+  charm:
+    charm-binary-python-packages:
+      - jsonschema
```
nsklikas, line 85: For some reason packing the charm fails without this. Not sure what changed.

natalian98 (approved):

PR #176, opened 2024-08-05
Fix CI

natalian98:


BarcoMasile (approved):

nsklikas:
So the issue was the juju version?

natalian98:
> So the issue was the juju version?

The issue was with not trusting traefik. I also bumped the version since we pinned to 3.1.

nsklikas:
> > So the issue was the juju version?
> 
> The issue was with not trusting traefik. I also bumped the version since we pinned to 3.1.

But the message in the failed run is `Waiting for database creation`.

Also the ci now passes in main, which looks like a transient error (which we need to look at). Am I missing something?

natalian98:
The CI in main passed, but another recent run failed, see https://github.com/canonical/hydra-operator/actions/runs/10256853322/job/28376804392.
This seems to be a transient error indeed, as postgresql-k8s was not recently updated.

Nonetheless, we should trust traefik as it needs to be able to patch k8s services.

nsklikas (approved):

PR #231, opened 2024-11-27
feat: replace traefik with istio-ingress for public ingress

wood-push-melon:
This pull request aims to drop traefik-k8s and use istio-ingress-k8s charm for public ingress.

Note:
- For internal ingress, we need the [feature support](https://github.com/canonical/istio-ingress-k8s-operator/issues/23) in istio-ingress-k8s. Therefore, we are still using the traefik-k8s charm for it. We'll come back to it when istio-ingress-k8s charm supports that.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -31,7 +31,8 @@ jobs:
         with:
           provider: microk8s
           channel: 1.28-strict/stable
-          juju-channel: 3.4
+          juju-channel: 3.4/stable
+          microk8s-addons: "hostpath-storage metallb:10.64.140.43-10.64.140.49"
 
       - name: Run integration tests
         # set a predictable model name so it can be consumed by charm-logdump-action
--- a/README.md
+++ b/README.md
@@ -17,9 +17,16 @@ visit <https://www.ory.sh/docs/hydra/>.
 ## Usage
 
 ```shell
+# Deploy required charms
 juju deploy postgresql-k8s --channel edge --trust
+juju deploy istio-ingresss-k8s public-ingress --channel latest/edge --trust
+
+# Deploy hydra
 juju deploy hydra --trust
-juju integrate postgresql-k8s hydra
+
+# Integrate with required charms
+juju integrate hydra postgresql-k8s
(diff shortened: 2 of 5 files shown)


nsklikas (approved):
So no charm changes were required? Cool.

For the internal ingress my understanding was that we will drop that and use the mesh interface instead, or maybe I misunderstood?

@wood-push-melon @shipperizer should we keep tests for traefik to be sure that we don't break that integration or we no longer care about that??

wood-push-melon:
> So no charm changes were required? Cool.
> 
> For the internal ingress my understanding was that we will drop that and use the mesh interface instead, or maybe I misunderstood?
> 
> @wood-push-melon @shipperizer should we keep tests for traefik to be sure that we don't break that integration or we no longer care about that??

Yeah, no charm changes required generally. The `ingress` interface is reused.

(I might be wrong too) For the internal ingress, what I understand is that (at the current stage) we still need "something" to let the HTTPRoutes attach to, so that the traffic coming from other k8s clusters will get routed to different hydra listening ports. And this "something" will be another istio ingress (Gateway) at the current point. I saw there is some initiative called [GAMMA](https://gateway-api.sigs.k8s.io/mesh/gamma/), which lets us bind HTTPRoutes to k8s svc. It looks like an ongoing development.

I think we will no longer care about traefik in the future. But happy to hand it over to you two's decisions :)

shipperizer:
> So no charm changes were required? Cool.
> 
> For the internal ingress my understanding was that we will drop that and use the mesh interface instead, or maybe I misunderstood?
> 
> @wood-push-melon @shipperizer should we keep tests for traefik to be sure that we don't break that integration or we no longer care about that??

as discussed at standup, let s drop the `internal` ingress and use the service mesh

wood-push-melon:
> > So no charm changes were required? Cool.
> > For the internal ingress my understanding was that we will drop that and use the mesh interface instead, or maybe I misunderstood?
> > @wood-push-melon @shipperizer should we keep tests for traefik to be sure that we don't break that integration or we no longer care about that??
> 
> as discussed at standup, let s drop the `internal` ingress and use the service mesh

yep, will make it in another PR.

PR #235, opened 2024-12-05
feat: drop internal ingress

wood-push-melon:
This pull request aims to drop the internal ingress.


tests/integration/test_charm.py
```diff
@@ -107,13 +107,14 @@ async def test_build_and_deploy(ops_test: OpsTest, local_charm: Path) -> None:
...
     await ops_test.model.wait_for_idle(
-        apps=[HYDRA_APP, DB_APP, PUBLIC_INGRESS_APP, INTERNAL_INGRESS_APP],
+        apps=[HYDRA_APP, DB_APP, PUBLIC_INGRESS_APP, ADMIN_INGRESS_APP],
         raise_on_blocked=False,
         status="active",
         timeout=5 * 60,
     )
 
 
+@pytest.mark.skip
```
pik4ez-canonical, line 117: What is the reason to skip these tests?

pik4ez-canonical (changes requested):
I'm curious about a bunch of skipped tests.

src/charm.py
```diff
@@ -187,28 +188,10 @@ def __init__(self, *args: Any) -> None:
             self._on_database_integration_broken,
         )
 
-        # admin ingress
-        self.framework.observe(self.admin_ingress.on.ready, self._on_admin_ingress_ready)
```
natalian98, line 191: Why is this removed? Don't we still need to observe the admin ingress-related events like we continue for the public one?

tests/integration/test_charm.py
wood-push-melon: ah, forgot to remove those when I did debugging. Let me remove those.

src/charm.py
wood-push-melon: this is because we don't do anything for admin-ingress integration joining and breaking.

pik4ez-canonical (approved):

BarcoMasile (approved):

src/charm.py
wood-push-melon: what I understand is that:

- We integrate with an admin ingress to expose admin API endpoints if we would like some client (outside the k8s cluster) to query
- For integration `hydra_endpoints`, we only pass the k8s svc addresses which are fixed values and for internal-cluster usage.

So admin ingress integration joining/departing does not change anything.

Let me know if I am wrong (we can chat a bit tomorrow)

natalian98: Ok, I missed this part:

> For integration hydra_endpoints, we only pass the k8s svc addresses which are fixed values and for internal-cluster usage.

It's fine to remove it then.

natalian98 (approved):

PR #250, opened 2025-01-14
CI: deployment via JIMM

shipperizer:
deploying to our environments using GH actions and JIMM

https://github.com/canonical/hydra-operator/actions/workflows/publish.yaml

### DIFF
--- a/.github/workflows/_deploy.yaml
+++ b/.github/workflows/_deploy.yaml
@@ -1,0 +1,55 @@
+# reusable workflow triggered manually
+name: Deploy via JIMM
+
+on:
+  workflow_call:
+    inputs:
+        model:
+            type: string
+        revision:
+            type: string
+        channel:
+          type: string
+    secrets:
+        CLIENT_ID:
+            required: true
+        CLIENT_SECRET:
+            required: true
+        JIMM_URL:
+          required: true
+
+jobs:
+  deploy:
+    name: Deploy
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


.github/workflows/_deploy.yaml
BarcoMasile, line 1: small suggestion: should we call it something like "jimm_deploy" or "perform_deployment" so it stands out more from the other deploy.yaml?

deployment/main.tf
```diff
@@ -0,0 +1,77 @@
...
+  type        = string
+}
+
+variable "client_id" {
+  type        = string
+}
+
+variable "client_secret" {
+  type        = string
+}
```
BarcoMasile, line 20: suggestion: add `sensitive   = true` at least for this one var

.github/workflows/publish.yaml
```diff
@@ -57,11 +70,73 @@ jobs:
...
+    runs-on: ubuntu-24.04
+    if: ${{ inputs.source_branch == 'IAM-1233' }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+        with:
+          fetch-depth: 0
+          ref: ${{ inputs.source_branch }}
+      - name: Get revision
+        id: set-revision
```
BarcoMasile, line 98: This hurts my whole body

shipperizer: fixed

BarcoMasile (approved):

natalian98 (approved):
lgtm, let's merge the PRs for other charms once we make sure the hydra CI works well

PR #282, opened 2025-03-11
Do not remove client on relation_broken

### LINKED ISSUE #268: Investigate client deleted on scale down
### Bug Description

We have noticed that sometimes the client_id that is in the databag may be deleted from the hydra database. Also we are occasionally getting this error:
```
unit-grafana-0: 16:10:53 ERROR unit.grafana/0.juju-log grafana-source:14: Uncaught exception while in charm code:
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/model.py", line 3320, in _run
    result = subprocess.run(args, **kwargs)  # type: ignore
  File "/usr/lib/python3.8/subprocess.py", line 516, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '('/var/lib/juju/tools/unit-grafana-0/secret-get', 'secret://32140fb1-ca8d-4581-8700-6556654e47f5/curgiunmp25c77sv7i0g', '--format=json')' returned non-zero exit status 1.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "./src/charm.py", line 1636, in <module>
    main(GrafanaCharm, use_juju_for_storage=True)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/__init__.py", line 343, in __call__
    return _main.main(charm_class=charm_class, use_juju_for_storage=use_juju_for_storage)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/_main.py", line 543, in main
    manager.run()
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/_main.py", line 529, in run
    self._emit()
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/_main.py", line 518, in _emit
    _emit_charm_event(self.charm, self.dispatcher.event_name, self._juju_context)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/_main.py", line 134, in _emit_charm_event
    event_to_emit.emit(*args, **kwargs)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/framework.py", line 347, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/framework.py", line 857, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/framework.py", line 947, in _reemit
    custom_handler(event)
  File "/var/lib/juju/agents/unit-grafana-0/charm/lib/charms/tempo_coordinator_k8s/v0/charm_tracing.py", line 735, in wrapped_function
    return callable(*args, **kwargs)  # type: ignore
  File "/var/lib/juju/agents/unit-grafana-0/charm/lib/charms/grafana_k8s/v0/grafana_source.py", line 566, in _on_grafana_source_relation_changed
    self.on.sources_changed.emit()  # pyright: ignore
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/framework.py", line 347, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/framework.py", line 857, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/framework.py", line 947, in _reemit
    custom_handler(event)
  File "/var/lib/juju/agents/unit-grafana-0/charm/lib/charms/tempo_coordinator_k8s/v0/charm_tracing.py", line 735, in wrapped_function
    return callable(*args, **kwargs)  # type: ignore
  File "./src/charm.py", line 438, in _on_grafana_source_changed
    self._configure()
  File "/var/lib/juju/agents/unit-grafana-0/charm/lib/charms/tempo_coordinator_k8s/v0/charm_tracing.py", line 735, in wrapped_function
    return callable(*args, **kwargs)  # type: ignore
  File "./src/charm.py", line 538, in _configure
    if self.containers["workload"].get_plan().services != self._build_layer().services:
  File "/var/lib/juju/agents/unit-grafana-0/charm/lib/charms/tempo_coordinator_k8s/v0/charm_tracing.py", line 735, in wrapped_function
    return callable(*args, **kwargs)  # type: ignore
  File "./src/charm.py", line 1087, in _build_layer
    oauth_provider_info = self.oauth.get_provider_info()
  File "/var/lib/juju/agents/unit-grafana-0/charm/lib/charms/hydra/v0/oauth.py", line 477, in get_provider_info
    _client_secret = self.get_client_secret(client_secret_id)
  File "/var/lib/juju/agents/unit-grafana-0/charm/lib/charms/hydra/v0/oauth.py", line 486, in get_client_secret
    client_secret = self.model.get_secret(id=client_secret_id)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/model.py", line 308, in get_secret
    content = self._backend.secret_get(id=id, label=label)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/model.py", line 3692, in secret_get
    result = self._run('secret-get', *args, return_output=True, use_json=True)
  File "/var/lib/juju/agents/unit-grafana-0/charm/venv/ops/model.py", line 3322, in _run
    raise ModelError(e.stderr) from e
ops.model.ModelError: ERROR permission denied

unit-grafana-0: 16:10:53 ERROR juju.worker.uniter.operation hook "grafana-source-relation-changed" (via hook dispatching script: dispatch) failed: exit status 1
unit-grafana-0: 16:10:53 INFO juju.worker.uniter awaiting error resolution for "relation-changed" hook
```

The error is probably related to scaling down hydra or the remote app.

nsklikas:
IAM-1344

There is no way to tell what caused the event to be emitted, it could be that the relation is departing but it could also mean that one of the applications is being scaled down.

Closes #268

### DIFF
--- a/lib/charms/hydra/v0/oauth.py
+++ b/lib/charms/hydra/v0/oauth.py
@@ -67,7 +67,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 10
+LIBPATCH = 11
 
 PYDEPS = ["jsonschema"]
 
@@ -419,12 +419,8 @@ class OAuthRequirer(OAuthRelation):
             self.on.invalid_client_config.emit(e.args[0])
 
     def _on_relation_broken_event(self, event: RelationBrokenEvent) -> None:
-        # Workaround for https://github.com/canonical/operator/issues/888
-        self._pop_relation_data(event.relation.id)
-        if self.is_client_created():
-            event.defer()
-            logger.info("Relation data still available. Deferring the event")
-            return
+        # This may be caused by a provider unit being removed.
+        # Also the oauth data may still be there, perhaps we should remove this event altogether for now.
 
         # Notify the requirer that the relation data was removed
         self.on.oauth_info_removed.emit()
@@ -734,10 +730,12 @@ class OAuthProvider(OAuthRelation):
         return f"client_secret_{relation.id}"
 
     def _on_relation_broken(self, event: RelationBrokenEvent) -> None:
+        # There is no way to tell if this event was emitted because the relation was removed or if one of
+        # the applications was scaled down. Until this is fixed, we don't delete the client.
         # Workaround for https://github.com/canonical/operator/issues/888
-        self._pop_relation_data(event.relation.id)
+        # self._pop_relation_data(event.relation.id)
 
-        self._delete_juju_secret(event.relation)
+        # self._delete_juju_secret(event.relation)
         self.on.client_deleted.emit(event.relation.id)
 
     def _create_juju_secret(self, client_secret: str, relation: Relation) -> Secret:
@@ -754,6 +752,9 @@ class OAuthProvider(OAuthRelation):
             return
         else:
             secret.remove_all_revisions()
+
+    def remove_secret(self, relation: Relation) -> None:
+        return self._delete_juju_secret(relation)
 
     def set_provider_info_in_relation_data(
         self,
--- a/src/charm.py
+++ b/src/charm.py
@@ -17,12 +17,7 @@ from charms.data_platform_libs.v0.data_interfaces import (
 )
 from charms.grafana_k8s.v0.grafana_dashboard import GrafanaDashboardProvider
 from charms.hydra.v0.hydra_endpoints import HydraEndpointsProvider
-from charms.hydra.v0.oauth import (
-    ClientChangedEvent,
-    ClientCreatedEvent,
-    ClientDeletedEvent,
-    OAuthProvider,
-)
+from charms.hydra.v0.oauth import ClientChangedEvent, ClientCreatedEvent, OAuthProvider
 from charms.identity_platform_login_ui_operator.v0.login_ui_endpoints import (
     LoginUIEndpointsRequirer,
 )
@@ -63,6 +58,7 @@ from constants import (
     INTERNAL_INGRESS_INTEGRATION_NAME,
     LOGGING_RELATION_NAME,
     LOGIN_UI_INTEGRATION_NAME,
+    OAUTH_INTEGRATION_NAME,
     PEER_INTEGRATION_NAME,
     PROMETHEUS_SCRAPE_INTEGRATION_NAME,
     PUBLIC_INGRESS_INTEGRATION_NAME,
@@ -72,7 +68,12 @@ from constants import (
     TEMPO_TRACING_INTEGRATION_NAME,
     WORKLOAD_CONTAINER,
 )
-from exceptions import MigrationError, PebbleServiceError
+from exceptions import (
+    ClientDoesNotExistError,
+    CommandExecError,
+    MigrationError,
+    PebbleServiceError,
+)
 from integrations import (
     DatabaseConfig,
     InternalIngressData,
@@ -173,6 +174,7 @@ class HydraCharm(CharmBase):
         )
 
         self.framework.observe(self.on.hydra_pebble_ready, self._on_hydra_pebble_ready)
+        self.framework.observe(self.on.update_status, self._holistic_handler)
         self.framework.observe(self.on.leader_elected, self._on_leader_elected)
         self.framework.observe(self.on.config_changed, self._on_config_changed)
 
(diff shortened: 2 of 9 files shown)


lib/charms/hydra/v0/oauth.py
```diff
@@ -733,12 +729,14 @@ def _get_client_config_from_relation_data(self, event: RelationChangedEvent) ->
     def _get_secret_label(self, relation: Relation) -> str:
         return f"client_secret_{relation.id}"
 
-    def _on_relation_broken(self, event: RelationBrokenEvent) -> None:
-        # Workaround for https://github.com/canonical/operator/issues/888
-        self._pop_relation_data(event.relation.id)
+    # There is no way to tell if this event was emittede because the relation was removed or if one of
+    # the applications was scaled down. Until this is fixed, we don't delete the client.
```
natalian98, line 733: Question: when scaled down, does only the departing unit receive this event? If that's the case maybe we could just check for leadership before deleting the client?

nsklikas: The docs have changed and I am struggling to find it, but afaict only the departing unit gets the event. Still if someone deletes the leader unit, all clients would be deleted. Even if juju will not do this on it's own, someone could do remove that unit manually. Not sure if we are fine with this.

BarcoMasile (approved):

nsklikas:
I added a reconciliation function that will deleted the client created by the `oauth` relation asynchronously. This will introduce some more load on the charm, how fast this will be depends on juju and how log it will take to get the peer relation data. Please have another go at reviewing it.

src/charm.py
```diff
@@ -671,6 +662,30 @@ def _on_rotate_key_action(self, event: ActionEvent) -> None:
...
+                self._cli.delete_oauth_client(client["client_id"])
+            except CommandExecError:
+                logger.error(
+                    f"Failed to delete the OAuth client bound with the oauth integration: {rel_id}"
+                )
+                continue
+            except ClientDoesNotExistError:
+                pass
+
+            self.peer_data.pop(k)
```
natalian98, line 687: Should we also delete the juju secret?

natalian98 (dismissed):
lgtm

BarcoMasile (dismissed):

src/charm.py
```diff
@@ -671,6 +662,30 @@ def _on_rotate_key_action(self, event: ActionEvent) -> None:
...
         event.set_results({"new-key-id": jwk_id})
 
+    @leader_unit
+    def _clean_up_oauth_relation_clients(self):
+        for k in list(self.peer_data.keys()):
+            if not k.startswith("oauth_"):
+                continue
+
+            rel_id = k[len("oauth_") :]
+            if self.model.get_relation(OAUTH_INTEGRATION_NAME, relation_id=int(rel_id)).active:
```
natalian98, line 672: Question: Would it be enough to just run this logic on `client_deleted`?
According to https://ops.readthedocs.io/en/latest/reference/ops.html#ops.Relation.active this is only inactive on relation broken events

src/cli.py
```diff
@@ -294,7 +294,9 @@ def delete_oauth_client(self, client_id: str) -> Optional[str]:
             stdout = self._run_cmd(cmd)
         except Error as err:
             logger.error("Failed to delete an OAuth client: %s", err)
-            return None
+            if "Unable to locate the resource" in err.stderr:
```
wood-push-melon, line 297: A nitpick (feel free to ignore): We need to `except ExecError` here, otherwise the `error` may not have the attribute `stderr`.

src/charm.py
```diff
@@ -671,6 +662,30 @@ def _on_rotate_key_action(self, event: ActionEvent) -> None:
         event.log("Successfully rotated the JWK")
         event.set_results({"new-key-id": jwk_id})
 
+    @leader_unit
+    def _clean_up_oauth_relation_clients(self):
+        for k in list(self.peer_data.keys()):
```
wood-push-melon, line 667: Just a nitpick (feel free to ignore): `for k in self.peer_data.keys()` should be good enough. `KeysView` should be iterable.

wood-push-melon (dismissed):

src/charm.py
nsklikas: I am `pop`ing from the peer relation data inside the loop, this causes an exception because the iterable is changing during the iteration.

src/charm.py
nsklikas: Good catch, I think that this means that this shouldn't be run on `client_deleted`, as the relation will be departing then.

src/charm.py
wood-push-melon: ic,  feel free to ignore the comment if it's too picky.

We could collect the deleted clients (keys), then pop them at the end of the function. So it won't cause any trouble I guess.

An illustration:
```
keys_to_remove = []
for k in self.peer_data.keys():
    ...
    # after a client is deleted
    keys_to_remove.append(k)

for k in keys_to_remove:
    self.peer_data.pop(k)
```

nsklikas: Yes this would work, I feel there is no great difference between the 2 approaches. Probably this one uses less memory since you don't have to create the list. Will change it

tests/unit/test_actions.py
```diff
@@ -585,3 +586,84 @@ def test_when_action_succeeds(
...
+                "oauth_1": json.dumps({"client_id": "client_id"}),
+                "oauth_2": json.dumps({"client_id": "client_id"}),
+                "oauth_3": json.dumps({"client_id": "client_id"}),
+            },
+        )
+
+        harness.run_action("reconcile-oauth-clients")
+
+        mocked_cli.call_count == 3
+        assert "Successfully deleted 3 clients"
```
wood-push-melon, line 669: LGTM. A minor thing:

```python
assert mocked_cli.call_count == 3
assert "Successfully deleted 3 clients" in output.logs
```

wood-push-melon (dismissed):

BarcoMasile (dismissed):

natalian98 (approved):

PR #337, opened 2025-05-07
Block charm if not integrated with Login UI

### LINKED ISSUE #322: Block hydra when login UI relation is not ready
### Bug Description

We are often seeing deployments where hydra is in an active state, BUT the URLs are not configured properly (it is using the `default-url` domain). We should investigate whether we have a race condition. I think that the easiest way to prevent this would be to block/wait hydra until we have a valid set of URLs from the login UI.

nsklikas:
IAM-1468

Closes #322

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -88,9 +88,9 @@ from utils import (
     container_connectivity,
     database_integration_exists,
     leader_unit,
+    login_ui_integration_exists,
     peer_integration_exists,
     public_ingress_integration_exists,
-    login_ui_integration_exists,
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


wood-push-melon (dismissed):
LGTM. Just a minor thing:

I guess it would be better to add a test case similar to the following in integration test.

```python
async def test_remove_public_ingress_integration(
    ops_test: OpsTest, hydra_application: Application
) -> None:
    async with remove_integration(ops_test, TRAEFIK_PUBLIC_APP, PUBLIC_INGRESS_INTEGRATION_NAME):
        assert hydra_application.status == "blocked"
```

nsklikas:
> LGTM. Just a minor thing:
> 
> I guess it would be better to add a test case similar to the following in integration test.
> 
> ```python
> async def test_remove_public_ingress_integration(
>     ops_test: OpsTest, hydra_application: Application
> ) -> None:
>     async with remove_integration(ops_test, TRAEFIK_PUBLIC_APP, PUBLIC_INGRESS_INTEGRATION_NAME):
>         assert hydra_application.status == "blocked"
> ```

Added

wood-push-melon (approved):

BarcoMasile (approved):

PR #355, opened 2025-06-24
Add token hook integration

nsklikas:
IAM-1555

A matching PR will be created for the hook-service-operator the will implement the other side of this relation. Setting this to draft until that is ready and tested

### DIFF
--- a/lib/charms/hydra/v0/hydra_token_hook.py
+++ b/lib/charms/hydra/v0/hydra_token_hook.py
@@ -12,7 +12,8 @@ config.
 
 import enum
 import logging
-from typing import Annotated, Any, List, Optional, TypeVar, Union, get_args
+from functools import cached_property
+from typing import List, Optional
 
 from ops import (
     CharmBase,
@@ -25,14 +26,11 @@ from ops import (
     RelationCreatedEvent,
     RelationEvent,
 )
-from pydantic import BaseModel as _BaseModel
 from pydantic import (
-    BeforeValidator,
+    BaseModel,
+    ConfigDict,
     Field,
-    PlainSerializer,
-    ValidationInfo,
 )
-from pydantic_core import from_json
 
 LIBID = "b2e5e865f0bc43638f1e4a0a63e899a9"
 LIBAPI = 0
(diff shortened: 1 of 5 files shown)


lib/charms/hydra/v0/hydra_token_hook.py
```diff
@@ -0,0 +1,252 @@
...
+LIBPATCH = 1
+
+PYDEPS = ["pydantic"]
+
+INTEGRATION_NAME = "hydra-token-hook"
+INTERFACE_NAME = "hydra_token_hook"
+logger = logging.getLogger(__name__)
+
+
+class BaseModel(_BaseModel):
```
wood-push-melon, line 48: Have a question about the section before the `AuthType`:

A bit hard to understand what it does at the first glance. Is it possible to simplify it a bit? Do `field_validator` and `field_serializer` help?

src/integrations.py
```diff
@@ -164,6 +165,52 @@ def load(cls, requirer: LoginUIEndpointsRequirer) -> "LoginUIEndpointData":
...
+                "token_hook_auth_in": self.auth_in,
+            })
+        return r
+
+    @classmethod
+    def load(cls: "HydraHookData", requirer: HydraHookRequirer) -> "HydraHookData":
+        if not (is_ready := requirer.ready()):
+            return cls()
+
+        data = requirer.consume_relation_data()
```
wood-push-melon, line 200: A minor thing:

We need to check if it's `None` before proceeding.

src/integrations.py
```diff
@@ -164,6 +165,52 @@ def load(cls, requirer: LoginUIEndpointsRequirer) -> "LoginUIEndpointData":
...
+            r.update({
+                "token_hook_auth_type": self.auth_type,
+                "token_hook_auth_name": self.auth_name,
+                "token_hook_auth_value": self.auth_value,
+                "token_hook_auth_in": self.auth_in,
+            })
+        return r
+
+    @classmethod
+    def load(cls: "HydraHookData", requirer: HydraHookRequirer) -> "HydraHookData":
```
wood-push-melon, line 196: A cosmetic thing: don't need to annotate the `cls`.

src/integrations.py
nsklikas: The `ready` function will return false if there is no data in the databag. So we can be sure that `consume_relation_data` will not return None.

lib/charms/hydra/v0/hydra_token_hook.py
natalian98, line 1: question: If hydra is the requirer, shouldn't the library be provided by hook-service?
It works eitherway, just looks a bit counterintuitive.

nsklikas: Who the requirer/provider is, is decided based on who provides the service/information. Here, hydra requires the information provided by the hook service.
The library should be defined in a central place, based on who needs to use it. Hydra should be the owner of the library as it is the only constant in the relation, other charms may be defined that provide this functionality as well. I don't think there will be other charms that implement the hydra hook logic though.

PR #356, opened 2025-06-25
IAM-1504

wood-push-melon:


src/configs.py
```diff
@@ -25,13 +26,21 @@ class CharmConfig:
...
             "dev_mode": self._config["dev"],
             "log_level": self._config["log_level"],
             "access_token_strategy": "jwt" if self._config["jwt_access_tokens"] else "opaque",
         }
 
+    def to_env_vars(self) -> EnvVars:
+        return {
+            "DEV": "true" if self._config["dev"] else "false",
+        }
+
```
nsklikas, line 43: question: Not sure I get this, do we need to set the env var, the `dev` key in the config and the `--dev` flag in the CLI?

natalian98 (approved):

PR #368, opened 2025-07-04
Don't restart service if config didn't change

### LINKED ISSUE #361: Service gets restarted on update_status event
### Bug Description

Every time the holistic handler runs, we restart the service. This became an issue recently when we started running the holistic handler on the `update_status`. We should update the holistic handler to restart the service only if the config file OR the env vars have changed.

nsklikas:
IAM-1613

Closes #361

### DIFF
--- a/src/services.py
+++ b/src/services.py
@@ -4,12 +4,14 @@
 import logging
 from collections import ChainMap
 from pathlib import PurePath
+from typing import Optional
 
+from ops import StoredState
 from ops.model import Container, ModelError, Unit
 from ops.pebble import Layer, LayerDict
 
 from cli import CommandLine
-from configs import ConfigFileManager
+from configs import ConfigFile
 from constants import (
     ADMIN_PORT,
     CONFIG_FILE_NAME,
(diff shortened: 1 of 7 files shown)


src/services.py
```diff
@@ -101,11 +102,19 @@ def prepare_dir(self, path: str | PurePath) -> None:
     def push_config_file(self, content: str) -> None:
         self._container.push(CONFIG_FILE_NAME, content, make_dirs=True)
 
-    def plan(self, layer: Layer) -> None:
+    def _restart_service(self, config_manager: ConfigFileManager) -> None:
+        if config_manager.config_changed:
+            self._container.restart(WORKLOAD_CONTAINER)
+        elif not self._container.get_service(WORKLOAD_CONTAINER).is_running():
+            self._container.start(WORKLOAD_CONTAINER)
```
wood-push-melon, line 109: Just a question:

Does `replan` cover the case of `start`?

nsklikas: Only if `startup = enabled`. But I think that we have disabled startup because pebble would try to start the service when the config wasn't ready, which created unnecessary error logs.

src/services.py
```diff
@@ -101,11 +102,19 @@ def prepare_dir(self, path: str | PurePath) -> None:
     def push_config_file(self, content: str) -> None:
         self._container.push(CONFIG_FILE_NAME, content, make_dirs=True)
 
-    def plan(self, layer: Layer) -> None:
+    def _restart_service(self, config_manager: ConfigFileManager) -> None:
+        if config_manager.config_changed:
```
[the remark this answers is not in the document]
nsklikas: IMHO it is better to avoid multiple `return`s wherever possible. I will leave it as it is, unless you have a strong objection.

wood-push-melon (approved):

BarcoMasile (dismissed):

PR #371, opened 2025-07-07
Scale up dev/stg units

nsklikas:
As discussed, I am scaling up our dev/stg envs to make it more closely mimic production use deployments


wood-push-melon (dismissed):

wood-push-melon:
looks like the "problematic" code in ops was introduced even in `2.23.0`. If the integration test continues to fail, try to limit the version to `2.22.0`.

BarcoMasile (dismissed):

PR #401, opened 2025-09-15
Introduce secret management actions and config

### LINKED ISSUE #391: Introduce action for updating the system secret
### Enhancement Proposal

When deploying a new instance of hydra and connecting it with an existing database, hydra cannot read the contents of the jwks table and any of the tables that contain session_data as they have been encrypted with an older system secrete. Our solution right now is to drop the jwks table and rotate the keys, the session data remain corrupted and unusable.

We should implement an action that allows the user to provide a new system secret key, this action would rotate the system secret key (see [here](https://www.ory.sh/docs/hydra/self-hosted/secrets-key-rotation)) putting the one the user provided at the top. 
This action should be paired with one that would allow the user to get the system secret keys.

Bonus: we could add a `type` param in the action that would allow the user provide the `system` or the `cookies` key.

nsklikas:
IAM-1683

To test deploy CIP, get the hydra keys from the juju secrets:
```
juju show-secret --reveal systemsecret
juju show-secret --reveal cookiesecret
```
Remove Hydra and deploy from this branch, by providing the old keys via juju secrets in the hydra config (as explained in the readme). Integrate the new Hydra with CIP and try to access the jwks endpoint.

The reason I chose not to go with a pure config implementation, is that an incorrect key can corrupt the whole database and it would require manually editing it to fix. For this reason, I chose to limit the user's freedom in order not to allow unexpected behavior as much as possible.

Closes #391

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,96 @@ juju integrate hydra:ui-endpoint-info identity-platform-login-ui-operator:ui-end
 juju integrate identity-platform-login-ui-operator:hydra-endpoint-info hydra:hydra-endpoint-info
 ```
 
+## Run Hydra from backup
+
+When migrating an Ory Hydra instance—for example, to a new server or environment—you need to ensure the new instance can decrypt existing user sessions and data. Hydra relies on two crucial secrets for this:
+
+1. System Secret: Used to encrypt sensitive data stored in the database, such as session payloads and JSON Web Key Sets (JWKS).
+2. Cookie Secret: Used to encrypt and sign Hydra's cookies.
+
+If you restore Hydra from a database backup without using the original secrets, the new instance will generate its own, rendering the backed-up data unusable. The Charmed Hydra Operator provides several helper actions and configuration options to manage these secrets and enable seamless server migration.
+
+### Key Management Actions
+
+The operator includes two Juju actions for managing secrets on a running Hydra instance.
+
+#### get-secret-keys
+
+This action retrieves the current secret keys used by Hydra. It's essential for backing up secrets before a migration.
+
+```console
+# Get the system secret keys
+juju run hydra/0 get-secret-keys type=system
+
+# Get the cookie secret keys
+juju run hydra/0 get-secret-keys type=cookie
+```
+
+#### add-secret-key
+
+This action adds a new secret key to Hydra's configuration. This is useful for key rotation or for adding a key from a backup to an existing deployment.
+
+```console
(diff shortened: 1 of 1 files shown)


natalian98 (approved):

src/secret.py
```diff
@@ -66,3 +75,43 @@ def to_service_configs(self) -> ServiceConfigs:
...
+        return [secret for _, secret in sorted(secrets.items(), reverse=True)]
+
+    def add_secret_key(self, typ: str, key: str) -> None:
+        """Add a new secret key."""
+        secret_key, secret_label = {
+            SYSTEM_SECRET: (SYSTEM_SECRET_KEY, SYSTEM_SECRET_LABEL),
+            COOKIE_SECRET: (COOKIE_SECRET_KEY, COOKIE_SECRET_LABEL),
+        }[typ]
+        secrets = self._secrets[secret_label] or {}
+        secrets[
```
wood-push-melon, line 109: Sorry for the late review. Have a nonblocking question here: what is this for?

src/secret.py
```diff
@@ -66,3 +75,43 @@ def to_service_configs(self) -> ServiceConfigs:
     def is_ready(self) -> bool:
         values = self.values()
         return all(values) if values else False
+
+
+class HydraSecrets:
```
wood-push-melon, line 80: Have a gut feeling (if I'm understanding it correctly) that we could make the `Secrets` as a base that only provides mapping-like data behaviors without involving any biz logic and labels. And the `HydraSecrets` inherits from it to implement the biz logic.

We can update it in the future.

src/secret.py
wood-push-melon: Gotcha. Just one minor thing I think we need to consider: we'd better use the monotonic clock instead of the wall clock if we are using timestamps for ordering (something like using the `time.monotonic_ns()`). For the secret creation time, users can use the CLI to get the secret metadata managed by Juju.

BarcoMasile (approved):

src/secret.py
nsklikas: AFAIK monotonic time relies on the system not restarting, on a k8s env where pods get scaled up and down it shouldn't solve our problem. I switched it to use a UTC timestamp and fixed a bug on the ordering. I understand your concern is that we may experience some weird issues if the clock goes back, but I am not sure there is much we can do about that. Also it would require someone to create a secret while that is happening which is quite unlikely. If you have a better idea please let me know.

PR #410, opened 2025-10-06
feat: List all details except secrets in `list-oauth-clients` output

DanielArndt:


nsklikas (dismissed):

tests/integration/test_charm.py
```diff
@@ -186,10 +186,12 @@ async def test_list_oauth_clients(oauth_clients: dict[str, str]) -> None:
 
 
 async def test_get_client(hydra_unit: Unit, oauth_clients: dict[str, str]) -> None:
+    clients = json.loads(oauth_clients["clients"])
```
wood-push-melon: A minor nitpick: probably you could put this logic into the fixture itself instead of doing it in each test case.

Another thing is that we probably need to also change this test case due to the changes:

```python
async def test_list_oauth_clients(oauth_clients: dict[str, str]) -> None:
    assert len(oauth_clients) > 0
```

DanielArndt: https://github.com/canonical/hydra-operator/pull/410/commits/02791958833c0dd77fc0f62a18cde8184d55342c

wood-push-melon (approved):

PR #424, opened 2025-10-20
IAM-1402: fix internal route relation

shipperizer:


natalian98 (commented):
Now only the internal ingress is written to the databag, see https://github.com/canonical/hydra-operator/blob/c642793628ddd40e49114dfba0ed077d466919bb/src/charm.py#L526-L530

shipperizer:
> Now only the internal ingress is written to the databag, see https://github.com/canonical/hydra-operator/blob/c642793628ddd40e49114dfba0ed077d466919bb/src/charm.py#L526-L530

how was it before it was changed?any chance you can have a look at the git blame?

technically login UI should use internal addresses to hit hydra, unless we use it to redirect

PR #430, opened 2025-11-05
Stop service when database is gone

### LINKED ISSUE #418: Stop the service when database is gone
### Bug Description

When the database is gone, the service needs to be stopped. Otherwise the connection/session remains active

nsklikas:
IAM-1758

Closes #418

- Stop the service when the database is gone
- Introduce collect_status handler to simplify holistic handler function
- Introduce a fix for traefik_route, the bug was causing the charm to go into an error state when it was removed

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -341,6 +341,9 @@ class HydraCharm(CharmBase):
         return self.charm_config["dev"]
 
     def _initialize_secrets(self) -> None:
+        if not self.unit.is_leader():
+            return
+
         if not (system_secrets := self.charm_config.get_system_secret()):
             self.hydra_secrets.add_secret_key(SYSTEM_SECRET, token_hex(16))
         else:
@@ -373,7 +376,6 @@ class HydraCharm(CharmBase):
         self._on_oauth_integration_created(event)
 
     def _on_internal_ingress_joined(self, event: RelationJoinedEvent) -> None:
-        self.unit.status = MaintenanceStatus("Configuring resources")
         self._on_internal_ingress_changed(event)
 
     def _on_internal_ingress_changed(self, event: RelationEvent) -> None:
@@ -383,6 +385,13 @@ class HydraCharm(CharmBase):
         self.internal_ingress._relation = event.relation
 
         if not self.internal_ingress.is_ready():
+            return
+        if event.relation.app is None:
+            # We need to defer the event as this is not handled in the holistic handler
+            # TODO(nsklikas): move this to the holistic handler and remove defer
+            # TODO2(nsklikas): Fix this in traefik_route lib, this is a bug and the lib should handle
+            # this in the `is_ready` method, like it does for the Provider side.
+            event.defer()
             return
 
         if self.unit.is_leader():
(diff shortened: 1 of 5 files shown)


src/charm.py
```diff
@@ -451,6 +461,11 @@ def _on_database_changed(self, event: DatabaseEndpointsChangedEvent) -> None:
     def _on_database_integration_broken(self, event: RelationBrokenEvent) -> None:
         self.unit.status = MaintenanceStatus("Configuring resources")
         self._holistic_handler(event)
+        try:
+            self._pebble_service.stop()
+        except PebbleServiceError as e:
+            logger.error(f"Failed to start the service, please check the container logs: {e}")
```
natalian98:
```suggestion
            logger.error(f"Failed to stop the service, please check the container logs: {e}")
```

BarcoMasile (approved):

BarcoMasile (approved):

src/integrations.py
```diff
@@ -227,6 +227,9 @@ class InternalIngressData:
     admin_endpoint: URL
     config: dict = field(default_factory=dict)
 
+    def is_ready(self) -> bool:
```
wood-push-melon: A minor thing: should we remove this (seems it's not being used)?

And probably it's simpler if we do `return self.public_endpoint and self.admin_endpoint` if we wanna keep this method.

src/utils.py
```diff
@@ -2,17 +2,22 @@
...
     PUBLIC_ROUTE_INTEGRATION_NAME,
     WORKLOAD_CONTAINER,
 )
+from integrations import LoginUIEndpointData, PublicRouteData
+
+if TYPE_CHECKING:
+    from charm import HydraCharm
 
 CharmEventHandler = TypeVar("CharmEventHandler", bound=Callable[..., Any])
 Condition = Callable[[CharmBase], bool]
```
wood-push-melon: A minor thing (feel free to ignore):

To make the type checker happy, probably now we need to do something like:

```python
CharmType = TypeVar("CharmType", bound=CharmBase)
Condition = Callable[[CharmType], bool]
```

src/charm.py
```diff
@@ -448,6 +481,10 @@ def _on_database_changed(self, event: DatabaseEndpointsChangedEvent) -> None:
     def _on_database_integration_broken(self, event: RelationBrokenEvent) -> None:
         self.unit.status = MaintenanceStatus("Configuring resources")
         self._holistic_handler(event)
+        try:
+            self._pebble_service.stop()
```
wood-push-melon: Just some thoughts and nothing needs to change here (like what we discussed in the sprint):
1. We'd better understand what IS did before, and come back to visit this again
2. We should consider if this is required for all required (hard) dependencies

tests/integration/test_charm.py
```diff
@@ -81,7 +81,7 @@ async def test_build_and_deploy(ops_test: OpsTest, local_charm: Path) -> None:
 
     await ops_test.model.integrate(f"{TRAEFIK_PUBLIC_APP}:certificates", f"{CA_APP}:certificates")
     # await ops_test.model.integrate(TRAEFIK_PUBLIC_APP, f"{LOGIN_UI_APP}:public-route") # TODO @shipperizer change this once login-ui is g2g
```
wood-push-melon: Remove this commented line?

src/integrations.py
nsklikas: removed it

PR #487, opened 2026-01-20
Migrate unit tests to ops.testing

nsklikas:
IAM-1764

### DIFF
--- a/src/cli.py
+++ b/src/cli.py
@@ -56,7 +56,7 @@ def parse_kv_string(kv_str: str) -> dict[str, str]:
 
 
 class OAuthClient(BaseModel):
-    redirect_uris: Optional[list[str]] = Field(
+    redirect_uris: list[str] = Field(
         default_factory=list,
         validation_alias=AliasChoices("redirect-uris", "redirect_uris", "redirect_uri"),
         serialization_alias="redirect-uris",
@@ -106,7 +106,7 @@ class OAuthClient(BaseModel):
 
     @property
     def managed_by_integration(self) -> bool:
-        return "integration-id" in self.metadata
+        return bool(self.metadata) and "integration-id" in self.metadata
 
     @field_validator("redirect_uris", mode="before")
     @classmethod
@@ -360,7 +360,7 @@ class CommandLine:
             stdout = self._run_cmd(cmd)
         except ExecError as err:
             logger.error("Failed to delete an OAuth client: %s", err)
-            if "Unable to locate the resource" in err.stderr:
+            if err.stderr and "Unable to locate the resource" in err.stderr:
                 raise ClientDoesNotExistError()
             raise CommandExecError() from err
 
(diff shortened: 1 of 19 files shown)


src/cli.py
```diff
@@ -56,7 +56,7 @@ def parse_kv_string(kv_str: str) -> dict[str, str]:
 
 
 class OAuthClient(BaseModel):
-    redirect_uris: Optional[list[str]] = Field(
+    redirect_uris: Optional[list[str] | str] = Field(
```
wood-push-melon: Question: why do we change the type hint?

tests/unit/test_secret.py
```diff
@@ -1,114 +1,97 @@
...
+                "secrets": [
+                    Secret(
+                        owner="app",
+                        label=COOKIE_SECRET_LABEL,
+                        tracked_content={COOKIE_SECRET_KEY: "cookie"},
+                    )
+                ]
+            }
+        ],
+        indirect=True,
```
wood-push-melon: Another way to look at the unit tests: do we need the Scenario's help to test this single `Secret` class?

Can we create mocks and test the methods' logic? Like what we do in the peer data.

tests/unit/test_services.py
```diff
@@ -1,165 +1,99 @@
...
+    )
+    def test_get_version(self, charm: HydraCharm, state: dict, expected: str) -> None:
+        workload_service = WorkloadService(charm.unit)
+        assert workload_service.version == expected
+
+    def test_open_port(self, charm: HydraCharm) -> None:
+        workload_service = WorkloadService(charm.unit)
+        workload_service.open_port()
 
-    def test_set_version(self, mocked_unit: MagicMock, workload_service: WorkloadService) -> None:
```
wood-push-melon: Have a question: Why do we remove these test cases?

And the same as above, we should think if it's necessary to bring Scenario into the picture here.

tests/unit/test_actions.py
```diff
@@ -111,69 +101,70 @@ def mocked_cli(self, mocker: MockerFixture, mocked_oauth_client_config: dict) ->
...
+                {
+                    "redirect-uris": ["https://example.oidc.client/callback"],
+                    "token-endpoint-auth-method": "client_secret_basic",
+                },
+            ),
+            state,
         )
 
         mocked_cli.assert_called_once()
-        assert output.results["redirect-uris"] == [mocked_oauth_client_config["redirect_uri"]]
```
wood-push-melon: (rusty memory, might be wrong): Is there a way we could verify the action results using the state output?

tests/unit/test_actions.py
```diff
@@ -611,68 +582,65 @@ def mocked_cli(self, mocker: MockerFixture) -> MagicMock:
...
-        harness.set_leader(True)
-        harness.update_relation_data(
-            peer_integration,
-            "hydra",
-            {
-                "oauth_1": json.dumps({"client_id": "client_id"}),
-                "oauth_2": json.dumps({"client_id": "client_id"}),
-                "oauth_3": json.dumps({"client_id": "client_id"}),
-            },
+        peer_relation_ready,
```
wood-push-melon: Same as above.

PR #573, opened 2026-07-02
feat: add arm64 support

natalian98:


nsklikas (approved):

tests/integration/utils.py
```diff
@@ -22,6 +23,16 @@ def juju_model_factory(model_name: str) -> jubilant.Juju:
...
 
+    # Set model constraints dynamically based on native host architecture
+    arch_map = {
+        "aarch64": "arm64",
+        "arm64": "arm64",
+        "x86_64": "amd64",
+        "amd64": "amd64",
+    }
+    host_arch = arch_map.get(platform.machine().lower(), "amd64")
+    juju.cli("set-model-constraints", f"arch={host_arch}")
```
nsklikas, line 34: nit (feel free to disregard): I am not sure if this should be part of the tests, feels like something that should be already configured from the caller. 

Otherwise perhaps it would be cleaner if we moved it to a dedicated autoused fixture.

natalian98: I'd prefer to keep this explicit for now to guarantee the model matches the runner architecture

PR #604, opened 2026-08-11
fix(oauth): reconcile oauth clients holistically instead of deferring

### LINKED ISSUE #560: OAuth client not created when ingress hostname is unavailable at relation time
**Note: This issue was generated with AI assistance (GitHub Copilot) based on automated log analysis and triage.**
Filed by @canonical/solutions-qa

---

## Observed behaviour

When Hydra processes an OAuth relation change (requirer populates `redirect_uri` and `grant_types`) before the Traefik public-route integration has provided an `external_host`, Hydra logs the following error repeatedly and does not write `client_id` or `client_secret` back into the relation:

```
ERROR oauth:10: External hostname is not set on the ingress provider
```

Once the Traefik ingress hostname subsequently becomes available, the OAuth client credentials are never written to the relation. The requirer (in this case `juju-jimm-k8s`) remains stuck with only server-level metadata (`authorization_endpoint`, `token_endpoint`, `issuer_url`, `token_endpoint_auth_method`) in the relation, but no `client_id` / `client_secret`.

## Steps to reproduce

Deploy the following bundle simultaneously (all apps deployed together, no ordering):

```
juju-jimm-k8s (rev 110, 3/edge)
hydra (rev 395, latest/edge)
traefik-k8s
pgbouncer-k8s
postgresql-k8s
openfga-k8s
self-signed-certificates
```

Relate:
- `hydra:oauth ↔ juju-jimm-k8s:oauth`
- `hydra:public-route ↔ traefik-k8s:traefik-route`

## Observed log sequence

From `hydra-0.log` (Juju container-agent output):

```
2026-06-01T02:21:04Z [container-agent] ERROR juju-log oauth:10: External hostname is not set on the ingress provider
2026-06-01T02:21:09Z [container-agent] ERROR juju-log oauth:10: External hostname is not set on the ingress provider
2026-06-01T02:21:29Z [container-agent] ERROR juju-log oauth:10: External hostname is not set on the ingress provider
```

From `machine-lock.log` for `hydra/0`:

```
02:21:07  relation-changed (10; app: target)  held 7s
02:21:14  relation-joined (10; unit: target/0)  held 8s
02:21:32  relation-changed (10; unit: target/0)  held 6s
...
02:22:13  relation-changed (12; app: traefik-k8s)  held 8s   ← external_host first available
02:22:19  relation-joined (12; unit: traefik-k8s/0)  held 6s
02:22:26  relation-changed (12; unit: traefik-k8s/0)  held 7s
```

No further `relation-changed (10; ...)` hooks run after `traefik-k8s` provides its hostname. The `client_id` and `client_secret` fields are absent from the relation data for the lifetime of the deployment.

## Relation data at end of deployment (from Juju db-dump)

**Hydra app-level data (provider side):**
```yaml
authorization_endpoint: https://10.141.4.115/oauth2/auth
issuer_url: https://10.141.4.115
token_endpoint: https://10.141.4.115/oauth2/token
token_endpoint_auth_method: client_secret_post
```
(`client_id` and `client_secret` absent)

**JIMM app-level data (requirer side):**
```yaml
grant_types: '["authorization_code", "refresh_token", "urn:ietf:params:oauth:grant-type:device_code"]'
redirect_uri: https://jimm-test.local/auth/callback
```

## Impact

`juju-jimm-k8s` remains in `blocked: Waiting for OAuth relation` indefinitely. The `test_deploy` integration test times out after 15 minutes.

Test Observer link:
https://test-observer.canonical.com/#/charms/408070?testExecutionId=540157&testResultId=11346342

This failure is reproducible across multiple test executions of `juju-jimm-k8s` rev 110 (executions 540157, 540245, 540250, 540274, 540278 and others all show the same `application:blocked:Waiting for OAuth relation` status).

## Environment

- Juju: 3.6.23
- Kubernetes cloud (`juju:3/stable ubuntu:22.04 kubernetes`)
- Hydra: rev 395, `latest/edge`
- `juju-jimm-k8s`: rev 110, `3/edge`
- Traefik: deployed from bundle `integration/content-cache-k8s:nginx-proxy/nginx-route/juju-jimm-k8s:nginx-route`

### LINKED ISSUE #591: Hydra charm does not update existing OAuth client redirect_uris when relation data changes
### Bug Description

When an OAuth requirer charm (such as oauth2-proxy) updates its requested `redirect_uri` in the relation data (for example, switching from an http:// to an https:// scheme), the hydra charm does not synchronize this change.

The Juju relation data successfully updates to show the new URI, but querying the Hydra API via the charm (`juju run hydra/0 list-oauth-clients`) shows that the client is still registered with the old URI.

This causes OIDC login flows to fail with a `redirect_uri_mismatch` error, as the proxy requests the new URI but Hydra only expects the old one. Restarting the Hydra StatefulSet does not resolve the issue. The only workaround is to completely remove and recreate the relation, which forces the charm to drop and recreate the client.

### To Reproduce

1. Deploy the IAM stack (e.g., via the `kubeflow-ambient-iam` terraform solution).
2. Integrate a requirer like `oauth2-proxy` with the `oauth` endpoint on the hydra charm while the ingress is configured for standard HTTP.
3. Verify the client is created with an HTTP redirect URI using j`uju run -m <model> hydra/0 list-oauth-clients`.
4. Update the ingress/public URL configuration to use HTTPS.
5. Inspect the Juju relation data (`juju show-unit`) and verify the requirer charm updated the `redirect_uri` field in the relation data to https://....
6. Re-run `juju run -m <model> hydra/0 list-oauth-clients`.
7. Notice the bug: The client in Hydra's database still reflects the old http:// URI, despite the updated relation data.

### Environment

Canonical k8s 1.32-classic/stable (v1.32.13)
Juju 3.6/stable (3.6.27)
hydra latest/stable revision 396
Identity platform terraform module v1.1.1

### Relevant log output

```shell
# 1. Relation data shows the CORRECT updated URI (HTTPS)
  - relation-id: 7
    endpoint: oauth
    cross-model: true
    related-endpoint: oauth
    application-data:
      redirect_uri: https://ui.kubeflow.com/kubeflow-oauth2-proxy/oauth2/callback

# 2. But the Hydra charm's internal client list shows the OLD URI (HTTP)
$ juju run -m iam hydra/0 list-oauth-clients
clients: '[{"redirect-uris": ["http://ui.kubeflow.com/kubeflow-oauth2-proxy/oauth2/callback"], ... }]'

# curl output


curl -L -v -k https://ui.kubeflow.com

cache-control: private, no-cache, no-store, must-revalidate
< content-type: text/html; charset=utf-8
< date: Tue, 28 Jul 2026 07:55:45 GMT
< location: https://auth.kubeflow.com/ui/oidc_error?error=invalid_request&error_description=The+request+is+missing+a+required+parameter%2C+includes+an+invalid+parameter+value%2C+includes+a+parameter+more+than+once%2C+or+is+otherw
ise+malformed.+The+%27redirect_uri%27+parameter+does+not+match+any+of+the+OAuth+2.0+Client%27s+pre-registered+redirect+urls.
< vary: Origin
< content-length: 368
< 
* Ignoring the response-body
100   368  100   368    0     0  26396      0 --:--:-- --:--:-- --:--:-- 26396
* Connection #1 to host auth.kubeflow.com left intact
* Issue another request to this URL: 'https://auth.kubeflow.com/ui/oidc_error?error=invalid_request&error_description=The+request+is+missing+a+required+parameter%2C+includes+an+invalid+parameter+value%2C+includes+a+parameter+more
+than+once%2C+or+is+otherwise+malformed.+The+%27redirect_uri%27+parameter+does+not+match+any+of+the+OAuth+2.0+Client%27s+pre-registered+redirect+urls.'
* Found bundle for host: 0x57aa61d991f0 [can multiplex]
* Re-using existing connection with host auth.kubeflow.com
* [HTTP/2] [3] OPENED stream for https://auth.kubeflow.com/ui/oidc_error?error=invalid_request&error_description=The+request+is+missing+a+required+parameter%2C+includes+an+invalid+parameter+value%2C+includes+a+parameter+more+than
+once%2C+or+is+otherwise+malformed.+The+%27redirect_uri%27+parameter+does+not+match+any+of+the+OAuth+2.0+Client%27s+pre-registered+redirect+urls.
* [HTTP/2] [3] [:method: GET]
* [HTTP/2] [3] [:scheme: https]
* [HTTP/2] [3] [:authority: auth.kubeflow.com]
* [HTTP/2] [3] [:path: /ui/oidc_error?error=invalid_request&error_description=The+request+is+missing+a+required+parameter%2C+includes+an+invalid+parameter+value%2C+includes+a+parameter+more+than+once%2C+or+is+otherwise+malformed.
+The+%27redirect_uri%27+parameter+does+not+match+any+of+the+OAuth+2.0+Client%27s+pre-registered+redirect+urls.]
* [HTTP/2] [3] [user-agent: curl/8.5.0]
* [HTTP/2] [3] [accept: */*]
} [5 bytes data]
> GET /ui/oidc_error?error=invalid_request&error_description=The+request+is+missing+a+required+parameter%2C+includes+an+invalid+parameter+value%2C+includes+a+parameter+more+than+once%2C+or+is+otherwise+malformed.+The+%27redirect_
uri%27+parameter+does+not+match+any+of+the+OAuth+2.0+Client%27s+pre-registered+redirect+urls. HTTP/2
> Host: auth.kubeflow.com
> User-Agent: curl/8.5.0
> Accept: */*
> 
{ [5 bytes data]
< HTTP/2 200 
< accept-ranges: bytes
< cache-control: no-cache, no-store, must-revalidate
< content-security-policy: default-src 'self' data: https://assets.ubuntu.com; script-src 'self' https://auth.kubeflow.com/.well-known/webauthn.js; style-src 'self'
< content-type: text/html; charset=utf-8
< date: Tue, 28 Jul 2026 07:55:45 GMT
< permissions-policy: interest-cohort=()
< vary: Origin
< x-content-type-options: nosniff
< x-frame-options: SAMEORIGIN
< content-length: 1545
< 
{ [1545 bytes data]
100  1545  100  1545    0     0  97090      0 --:--:-- --:--:-- --:--:-- 97090
* Connection #1 to host auth.kubeflow.com left intact
<!DOCTYPE html><html><head><meta charSet="utf-8" data-next-head=""/><meta name="viewport" content="width=device-width" data-next-head=""/><link rel="preload" href="./_next/static/css/6104b5e76d58eedc.css" as="style"/><link rel="p
reload" href="./_next/static/css/9da8fb12399c58f6.css" as="style"/><link rel="stylesheet" href="./_next/static/css/6104b5e76d58eedc.css" data-n-g=""/><link rel="stylesheet" href="./_next/static/css/9da8fb12399c58f6.css"/><noscrip
t data-n-css=""></noscript><script defer="" noModule="" src="./_next/static/chunks/polyfills-42372ed130431b0a.js"></script><script src="./_next/static/chunks/webpack-a3cbce65dab03667.js" defer=""></script><script src="./_next/sta
tic/chunks/framework-b1e5f14688f9ffe6.js" defer=""></script><script src="./_next/static/chunks/main-1f47325dad18d83a.js" defer=""></script><script src="./_next/static/chunks/pages/_app-35192eba4bb1cb95.js" defer=""></script><scri
pt src="./_next/static/chunks/726-5e0c8d887a7b70ee.js" defer=""></script><script src="./_next/static/chunks/pages/oidc_error-9ec67dcb5b2d9a4c.js" defer=""></script><script src="./_next/static/l1CaNSBrsKlBFrn3JCNX0/_buildManifest.
js" defer=""></script><script src="./_next/static/l1CaNSBrsKlBFrn3JCNX0/_ssgManifest.js" defer=""></script></head><body><div id="__next"></div><script id="__NEXT_DATA__" type="application/json">{"props":{"pageProps":{}},"page":"/
oidc_error","query":{},"buildId":"l1CaNSBrsKlBFrn3JCNX0","assetPrefix":".","nextExport":true,"autoExport":true,"isFallback":false,"scriptLoader":[]}</script></body></html>
```

### Additional context

Attempting to force a sync by rolling out the Hydra StatefulSet (`kubectl rollout restart sts/hydra`) does not update the client.

The viable workaround is to drop and recreate the relation, which forces the relation-departed hook to delete the client, and the subsequent relation-joined hook to recreate it with the correct data:

```
juju remove-relation oauth-offer:oauth oauth2-proxy:oauth
# Wait for removal
juju integrate oauth-offer:oauth oauth2-proxy:oauth
```

nsklikas:
## What

Replaces the event-driven OAuth client lifecycle with an idempotent holistic reconcile.

Client creation and update now run from `_holistic_handler`, and every `oauth` relation event is routed there too, so there is a single path. Drift between the requirer databag and the client registered in Hydra is detected with a sha256 fingerprint of the target configuration, stored next to the client id in the peer databag — so a change is applied even when no event carrying it survived.

Fixes #560
Fixes #591

## Why

**#560** — a requirer publishing its client config before Hydra is ready never receives `client_id`/`client_secret`. Two independent causes:

1. `_on_oauth_integration_created` called `event.defer()` when the public route URL was missing. Deferred events live in the pod-local `.unit-state.db`, wiped when the pod is recreated, and Juju never re-fires `relation-created`/`relation-joined` afterwards.
2. `OAuthProvider._get_client_config_from_relation_data` returns without emitting anything when Hydra's own app databag is empty — exactly the state cause 1 produces.

**#591** — after a client exists, a requirer changing `redirect_uri` is not propagated. `client_changed` only fired from `oauth-relation-changed` for the leader; a single missed occurrence was never retried, and nothing re-read the requirer databag on any other event.

The root cause of both is the same: **the charm assumed a deferred event would eventually run.** The fix is not to avoid deferring, it is to stop depending on delivery at all — the reconcile re-reads the requirer databag from scratch on every holistic pass, so a lost event costs nothing.

## Changes

| Area | Change |
|---|---|
| `src/oauth.py` (new) | `OAuthReconciler` — per-relation create/update plus the garbage collection moved out of `HydraCharm`. Log-and-continue per relation: a failure must not discard the peer records already staged for the others, because the retry would then register a second Hydra client for each of them. |
| `src/charm.py` | `oauth` relation events, internal-route events and token-hook events all observe `_on_holistic_handler`. Its tail is `_update_oauth_provider_info()` / `garbage_collect()` / `_reconcile_oauth_clients()`. `_on_oauth_integration_created`, `_on_oauth_client_created`, `_on_oauth_client_changed`, `_on_token_hook_changed` and `_clean_up_oauth_relation_clients` are deleted. |
| `src/cli.py` | New `oauth_client_exists`, mirroring the discriminator `delete_oauth_client` already uses: `False` only on Hydra's "Unable to locate the resource", `None` when the check itself failed, `True` otherwise. |
| `src/integrations.py` | `PeerData.get_record` — narrowing the databag encoding belongs in the Layer 2 wrapper that owns it. |
| `src/utils.py` | `leader_unit` deleted (no callers). |
| `charmcraft.yaml` | `reconcile-oauth-clients` description updated; the action now creates, updates, re-registers and deletes. |
| `lib/charms/hydra/v0/oauth.py` | `LIBPATCH` 12 → 13. New event-free `get_client_config` and `get_client_secret`. `_create_juju_secret` tolerates an existing label. Requirer- and provider-side validation no longer raise out of the hook. The requirer reads the latest secret revision. |

### Behaviour changes worth reviewing

- **Provider info is published only once the charm is ready.** It sits past the readiness guards, so a blocked charm no longer advertises endpoints it cannot serve.
- **Existing deployments self-heal.** Peer records written before this change carry no `config_hash`, so the first reconcile after upgrade issues one `update` and records it. That is the fix for already-broken #591 deployments.
- **`garbage_collect` retains the peer record when the delete fails**, so a later `reconcile-oauth-clients` retries instead of orphaning the client.
- **`hydra update client` without `--secret` preserves the secret**, so the update path never invalidates the juju secret held by the requirer.
- **A failing reconcile leaves the unit `Active`.** The workload is healthy; one integration is degraded. The failure is logged at error level and retried on the next holistic pass. Raising instead would discard the peer records already committed for the integrations that succeeded earlier in the same pass.
- **`event.defer()` is reachable in the oauth path** via `_holistic_handler`'s `EVENT_DEFER_CONDITIONS`. That is deliberate and harmless now: recovery no longer depends on the deferred event being delivered.
- **No `relation-broken` handling.** Deleting on `oauth-relation-broken` would regress #268 / canonical/operator#888.

## Review

Three rounds of expert review, 20+ findings, all resolved. The ones that changed behaviour:

- The reconcile could never repair a client that vanished from Hydra, while this branch's own action description promised exactly that. `_update_client` now probes `oauth_client_exists` and re-registers when the client is definitely gone; the action forces a pass rather than trusting the fingerprint.
- **Re-registering cut a new juju secret revision**, and the requirer reads the revision it tracks — so the repair silently left the requirer authenticating with a secret Hydra had discarded. It now re-registers with the secret the requirer already holds.
- `_create_client` published credentials *after* recording the peer entry; with per-relation exception handling that committed a fingerprint matching for ever while the requirer got nothing. Publish now happens first.
- The action could publish credentials into a databag carrying no provider info, and the next `oauth-relation-changed` then died in the library's unguarded provider-side `_load_data`, wedging both charms.
- The action reported success when it had reconciled nothing.

### Unit

`tox -e unit`: **205 passed, 5 xfailed, 1 xpassed** (from 182 on `main`). `tox -e lint` clean; mypy unchanged from baseline.

`tests/unit/test_oauth.py` is new. Run against a pristine `main`, its two headline tests fail — `create_oauth_client` and `update_oauth_client` "Called 0 times" for #560 and #591 respectively.

Every behavioural fix is **mutation-verified**: the defect is reintroduced one at a time and the named test must fail. 22 mutations, 22 caught, including publish-ordering, the leader guard, the existence discriminator, the forced action pass, and fingerprint stability.

### End-to-end on microk8s

Deployed in the #560 ordering — oauth requirer integrated **before** the public route.

**#560.** Workload stopped, client deleted, peer record wiped, pod deleted. Last oauth hook `16:43:38`; pod deleted `16:46:01`; afterwards only `start` / `hydra-pebble-ready` / `secret-changed` / `hydra-relation-changed` / `update-status` ran — **no oauth hook** — and the client was registered, the peer record restored and the requirer re-credentialed.

**#591.** Workload stopped, requirer's `redirect_uri` changed. The `oauth-relation-changed` carrying it was consumed and skipped (`Hydra service is not running, skipping the OAuth client reconciliation`), then Hydra's client picked up `https://10.64.140.43/hydra-oauth-test-oauth2-proxy-k8s/oauth2/callback` anyway, from a holistic hook.

**The secret-preservation fix.** Client deleted out of band, `reconcile-oauth-clients` run. New client registered, juju secret **revision unchanged at 3**, and the requirer's original secret still authenticates against the new client — `invalid_grant` (auth accepted, bogus code) rather than `invalid_client`, with a wrong-secret control returning `invalid_client`. The full OIDC flow completes: proxy `/oauth2/start` → Hydra `/oauth2/auth` → `302` login challenge.

**Discriminator.** Real Hydra puts `"error": "Unable to locate the resource"` on **stderr** with **exit 1** and an empty stdout, so `oauth_client_exists` distinguishes a missing client from an unreachable one.

Stack left at six applications `active/idle`.

## Upgrade path

The peer record gains a `config_hash` field inside the existing `oauth_<relation id>` key; no key is added or renamed. A pre-upgrade record is `{"client_id": ...}`, so `record.get("config_hash")` is `None`, never matches the fingerprint, and the first reconcile after upgrade issues **one** `hydra update client` against the existing client and records the hash. That is also the fix for deployments already broken by #591.

Validated on a live model: downgraded to `origin/main`, rewrote the peer record to the old shape, then refreshed to this branch.

```
peer record before : {"client_id": "07ce4967-..."}
peer record after  : {"client_id": "07ce4967-...", "config_hash": "c48d5b5e..."}
hooks after upgrade: upgrade-charm, config-changed, start, hydra-pebble-ready, hydra-relation-changed x3
hydra calls        : 1 update client (targeting 07ce4967-...), 0 create, 0 delete
juju secret        : revision 4 -> 4, value unchanged
```

The client id is preserved, no credentials are republished, and the requirer's pre-upgrade secret still authenticates against the post-upgrade client (`invalid_grant`, not `invalid_client`), with the full OIDC flow returning 302. Two further holistic passes issued zero Hydra calls, so it settles immediately.

## Notes

- `charmcraft pack` from a repo root containing a local `venv/` fails on a parts collision. Pre-existing; CI's clean checkout builds fine.
- A credential publish that keeps failing leaks one Hydra client per reconcile pass. De-duplicating would mean listing every client and filtering on `metadata.integration-id` on every pass — a deliberate non-goal; the failure is logged throughout.
- `LIBPATCH` is bumped to 13; publishing to Charmhub can happen separately, since `lib/` is bundled at build time.

### DIFF
--- a/src/cli.py
+++ b/src/cli.py
@@ -266,7 +266,14 @@ class CommandLine:
     def get_oauth_client(self, client_id: str) -> Optional[OAuthClient]:
         """Get an OAuth 2.0 client by client id.
 
+        Returns None when the client could not be fetched. A caller that must not confuse
+        "Hydra does not have it" with "the lookup failed" should catch the exception rather
+        than test for None.
+
         More information: https://www.ory.sh/docs/hydra/cli/hydra-get-client
+
+        Raises:
+            ClientDoesNotExistError: if Hydra reports that the client is not registered.
         """
         cmd = [
             "hydra",
@@ -281,10 +288,13 @@ class CommandLine:
 
         try:
             stdout = self._run_cmd(cmd)
+        except ExecError as err:
+            logger.error("Failed to get the OAuth client: %s", err)
+            if err.stderr and "Unable to locate the resource" in err.stderr:
+                raise ClientDoesNotExistError() from err
+            return None
         except Error as err:
             logger.error("Failed to get the OAuth client: %s", err)
-            if "Unable to locate the resource" in str(err):
-                logger.error("OAuth client not found: %s", client_id)
             return None
 
         return OAuthClient.model_validate_json(stdout)
--- a/src/oauth.py
+++ b/src/oauth.py
@@ -1,0 +1,229 @@
+# Copyright 2026 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Reconciliation between the `oauth` integrations and the OAuth2 clients in Hydra."""
+
+import hashlib
+import json
+import logging
+from typing import Any
+
+from charms.hydra.v0.oauth import DataValidationError, OAuthProvider
+from ops.model import Model, Relation
+
+from cli import CommandLine, OAuthClient
+from constants import OAUTH_INTEGRATION_NAME
+from exceptions import ClientDoesNotExistError, CommandExecError
+from integrations import PeerData
+
+logger = logging.getLogger(__name__)
+
+OAUTH_PEER_KEY_PREFIX = "oauth_"
+
+
+def _fingerprint(client: OAuthClient) -> str:
+    """Return a stable hash of the client configuration Hydra is expected to hold.
+
+    The client id and secret are excluded so that the create pass and every later update
+    pass hash identically, and so that no digest of a secret is ever written to the peer
+    databag. List values are sorted because their order carries no meaning to Hydra and a
+    requirer that derives them from a set would otherwise produce a new digest every hook.
+    """
+    payload = client.model_dump(
+        by_alias=True, exclude_none=True, exclude={"client_id", "client_secret"}, mode="json"
+    )
+    canonical = {k: sorted(v) if isinstance(v, list) else v for k, v in payload.items()}
+    return hashlib.sha256(json.dumps(canonical, sort_keys=True).encode()).hexdigest()
+
+
+class OAuthReconciler:
+    """Synchronize the `oauth` integrations with the OAuth2 clients registered in Hydra."""
+
+    def __init__(
+        self,
+        model: Model,
+        peer_data: PeerData,
+        oauth_provider: OAuthProvider,
+        cli: CommandLine,
+    ) -> None:
+        self._model = model
+        self._peer_data = peer_data
+        self._oauth_provider = oauth_provider
+        self._cli = cli
+
+    def reconcile(self, force: bool = False) -> list[int]:
+        """Create or update the Hydra OAuth2 client of every `oauth` integration.
+
+        The requirer databag of every integration is read on every pass: drift can only be
+        detected by comparing the published configuration against the recorded fingerprint,
+        so the peer record cannot short-circuit the read.
+
+        Args:
+            force: reapply the configuration even when the fingerprint is unchanged. Used by
+                the `reconcile-oauth-clients` action to repair clients that were changed or
+                removed in Hydra behind the charm's back.
+
+        Returns:
+            The ids of the integrations whose Hydra client could not be reconciled. An
+            integration whose requirer data is invalid is not among them: only the requirer
+            can fix that, and the next pass picks it up once they do.
+        """
+        if not self._model.unit.is_leader():
+            return []
+
+        failed: list[int] = []
+        for relation in self._model.relations[OAUTH_INTEGRATION_NAME]:
+            try:
+                if not self._reconcile_relation(relation, force):
+                    failed.append(relation.id)
+            except DataValidationError:
+                logger.warning(
+                    "The requirer data of the oauth integration %d is invalid, skipping it",
+                    relation.id,
+                )
+            except Exception:
+                # A failure here must not abort the hook: the peer records already staged for
+                # the integrations reconciled earlier in this pass would be discarded, and the
+                # retry would create a second Hydra client for each of them.
+                logger.exception("Failed to reconcile the oauth integration %d", relation.id)
+                failed.append(relation.id)
+
+        return failed
+
+    def garbage_collect(self) -> int:
+        """Delete the OAuth2 clients whose `oauth` integration no longer exists."""
+        if not self._model.unit.is_leader():
+            return 0
+
+        deleted: list[str] = []
+        for key in self._peer_data.keys():
+            if not key.startswith(OAUTH_PEER_KEY_PREFIX):
+                continue
+
+            relation_id = int(key[len(OAUTH_PEER_KEY_PREFIX) :])
+            relation = self._model.get_relation(OAUTH_INTEGRATION_NAME, relation_id=relation_id)
+            if relation is None or relation.active:
+                continue
+
+            record = self._record(relation_id)
+            if not (client_id := record.get("client_id")):
+                logger.warning(
+                    "The peer data of the oauth integration %d holds no client id, dropping it",
+                    relation_id,
+                )
+                deleted.append(key)
+                continue
+
+            try:
+                self._cli.delete_oauth_client(client_id)
+            except CommandExecError:
+                logger.error(
+                    "Failed to delete the OAuth client bound with the oauth integration: %d. "
+                    "Please run the 'reconcile-oauth-clients' action.",
+                    relation_id,
+                )
+                continue
+            except ClientDoesNotExistError:
+                pass
+
+            self._oauth_provider.remove_secret(relation)
+            deleted.append(key)
+
+        for key in deleted:
+            self._peer_data.pop(key)
+
+        return len(deleted)
+
+    def _peer_key(self, relation_id: int) -> str:
+        return f"{OAUTH_PEER_KEY_PREFIX}{relation_id}"
+
+    def _record(self, relation_id: int) -> dict[str, Any]:
+        """Read an integration's client record, treating a malformed value as absent."""
+        value = self._peer_data[self._peer_key(relation_id)]
+        return value if isinstance(value, dict) else {}
+
+    def _reconcile_relation(self, relation: Relation, force: bool) -> bool:
+        if not (client_config := self._oauth_provider.get_client_config(relation)):
+            logger.info(
+                "The oauth integration %d has no client configuration yet, skipping it",
+                relation.id,
+            )
+            return True
+
+        target = OAuthClient(
+            **client_config.to_dict(),
+            metadata={"integration-id": str(relation.id)},
+        )
+        fingerprint = _fingerprint(target)
+        record = self._record(relation.id)
+
+        if not record:
+            return self._create_client(relation, target, fingerprint)
+
+        if not (client_id := record.get("client_id")):
+            logger.error(
+                "The peer data of the oauth integration %d holds no client id, skipping it",
+                relation.id,
+            )
+            return False
+
+        if record.get("config_hash") == fingerprint and not force:
+            return True
+
+        return self._update_client(relation, target, client_id, fingerprint)
+
+    def _create_client(self, relation: Relation, target: OAuthClient, fingerprint: str) -> bool:
+        if not (created := self._cli.create_oauth_client(target)):
+            logger.error("Failed to create the OAuth client bound with the oauth integration")
+            return False
+
+        # Publish first: recording first would commit a fingerprint that matches for ever if
+        # the publish then failed, leaving the requirer without credentials permanently.
+        self._oauth_provider.set_client_credentials_in_relation_data(
+            relation.id,
+            created.client_id,  # type: ignore[arg-type]
+            created.client_secret,  # type: ignore[arg-type]
+        )
+        self._peer_data[self._peer_key(relation.id)] = {
+            "client_id": created.client_id,
+            "config_hash": fingerprint,
+        }
+        return True
+
+    def _update_client(
+        self, relation: Relation, target: OAuthClient, client_id: str, fingerprint: str
+    ) -> bool:
+        target.client_id = client_id
+        if self._cli.update_oauth_client(target):
+            self._peer_data[self._peer_key(relation.id)] = {
+                "client_id": client_id,
+                "config_hash": fingerprint,
+            }
+            return True
+
+        # Only a definite "Hydra does not have it" justifies registering a replacement; an
+        # inconclusive lookup would leave the original client live and unmanaged.
+        try:
+            self._cli.get_oauth_client(client_id)
+        except ClientDoesNotExistError:
+            pass
+        else:
+            logger.error(
+                "Failed to update the OAuth client bound with the oauth integration: %d",
+                relation.id,
+            )
+            return False
+
+        # The client is gone from Hydra (database restored, deleted out of band). The peer
+        # record is the only trace left, so register a replacement and republish it.
+        #
+        # Re-register with the secret the requirer already holds. Writing a different one
+        # would cut a new juju secret revision, and the requirer reads the revision it
+        # tracks, so it would keep authenticating with a secret Hydra no longer accepts.
+        logger.warning(
+            "The OAuth client of the oauth integration %d no longer exists in Hydra, recreating it",
+            relation.id,
+        )
+        target.client_id = None
+        target.client_secret = self._oauth_provider.get_client_secret(relation)
+        return self._create_client(relation, target, fingerprint)
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -277,7 +277,7 @@ actions:
         default: RS256
         enumerate: [RS256, RS512, ES256, ES512, EdDSA]
   reconcile-oauth-clients:
-    description: Clean up clients that were created by the oauth relation
+    description: Synchronize the oauth clients with the oauth integrations, creating or updating the clients of active integrations, re-registering clients that were removed from Hydra out of band, and deleting the clients of removed integrations
   get-secret-keys:
     description: Get all the system or cookie secret keys used by hydra to encrypt sensitive data
     params:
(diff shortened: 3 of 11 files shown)


shipperizer (dismissed):

src/integrations.py
```diff
@@ -47,6 +47,11 @@ def __getitem__(self, key: str) -> JsonSerializable:
         value = peers.data[self._app].get(key)
         return json.loads(value) if value else {}
 
+    def get_record(self, key: str) -> dict[str, Any]:
```
wood-push-melon: This seems a bit strange in the class `PeerData` which is supposed to be a dict-like class. Is it necessary to add this method?

src/cli.py
```diff
@@ -289,6 +289,40 @@ def get_oauth_client(self, client_id: str) -> Optional[OAuthClient]:
 
         return OAuthClient.model_validate_json(stdout)
 
+    def oauth_client_exists(self, client_id: str) -> Optional[bool]:
```
wood-push-melon: Is this method necessary? Can the caller use `get_oauth_client`?

src/oauth.py
```diff
@@ -0,0 +1,219 @@
...
+    requirer that derives them from a set would otherwise produce a new digest every hook.
+    """
+    payload = client.model_dump(
+        by_alias=True, exclude_none=True, exclude={"client_id", "client_secret"}, mode="json"
+    )
+    canonical = {k: sorted(v) if isinstance(v, list) else v for k, v in payload.items()}
+    return hashlib.sha256(json.dumps(canonical, sort_keys=True).encode()).hexdigest()
+
+
+class OAuthReconciler:
```
wood-push-melon, line 39: Nonblocking - some immature idea (could be wrong; we can think about this more):

1. The hook-triggered OAuth client operations sit in the charm. For example, oauth integration created, changed, broken etc.
2. The reconciliation logic sits in another pebble service (might act like a cron) to reconcile the OAuth clients

nsklikas: The main problem with this is that half the logic is on the juju side (the oauth relations). Probably we can leverage notices for this.

If I understand correctly, the charm would write clients that it expects to exist in a file (just dumps all the clients from the oauth relations) on the workload container and we would have a script that wakes up every N hours to make sure that the clients in the file actually exist. We will need to make sure that the script never runs before the file is created. Also the script would have to do the caching to a file (since it won't be able to access the peer relation) and we don't want to spam the hydra API.

I think that we can give it a shot, but I would push for this PR to be merged first.

wood-push-melon: Yeah, something like that to decouple the charm from the OAuth client management. And probably the action could also disappear.

wood-push-melon (approved):

